Issue #68, opened 2026-04-15
AOT compiler: register state not preserved across loop iterations and missing call codegen


cataggar:
## Summary

The AOT compiler produces incorrect results for loops that accumulate values across iterations. The register allocator assigns fresh VRegs per instruction without persisting register state across block boundaries, so values computed in one loop iteration are not available in the next.

## Reproduction

`wat
(module
  (func (export "_start") (result i32)
    (local i32 i32)  ;; i, sum
    i32.const 0 local.set 1  ;; sum = 0
    i32.const 1 local.set 0  ;; i = 1
    (block (loop
      local.get 0 local.get 1 i32.add local.set 1  ;; sum += i
      local.get 0 i32.const 1 i32.add local.set 0  ;; i += 1
      local.get 0 i32.const 51 i32.lt_s br_if 0    ;; loop while i < 51
    ))
    local.get 1))
`

`
wamrc loop.wasm -o loop.aot && wamr loop.aot
Expected: 1275  (sum 1..50)
Actual:   1
`

Wasmtime produces 1275 correctly. The WAMR interpreter also handles this correctly.

## Root Cause

Two related issues in the x86-64 codegen:

1. **Register allocator does not reset across blocks.** `RegMap.assign()` gives each VReg a fresh register. When a loop body is compiled, each iteration's `local_get` gets a new VReg and thus a new register, but the value stored by the previous iteration's `local_set` went to a stack frame slot `[rbp-N]`. The new `local_get` correctly emits `mov reg, [rbp-N]`, but the register allocator may have exhausted scratch registers from prior iterations' dead VRegs.

2. **Function calls (`call` op) are stubbed.** The codegen emits `mov dest, 0` instead of actually calling the target function. This blocks fibonacci and any multi-function program.

## Impact

- Any loop that accumulates (sum, product, counter)
- Recursive fibonacci (also needs `call` codegen)
- Real-world programs (virtually all use loops)

### Short-term: Register allocator improvements
- Reset register allocations at block boundaries (free dead VRegs)
- Or: switch to a simpler approach where ALL values live on the stack frame (locals-only, no register caching). This is slower but correct.

### Medium-term: Implement `call` codegen
- Compute target function address from function offset table
- Set up SysV ABI arguments (rdi, rsi, rdx, rcx, r8, r9)
- Emit `call rel32` with patch list (same pattern as branch patching)
- Collect return value from RAX

### Long-term: Proper register allocator
[...]

PR #69, opened 2026-04-15
feat(aot): stack-based codegen fixes loop accumulation + adds call support

cataggar:
## Summary

Replace the broken register allocator with a stack-based operand model, fixing loop accumulation and adding intra-module function call support.

### Problem
The RegMap register allocator never freed registers. After 9 VRegs, all values were silently dropped. Loops returned wrong results (1 instead of 1275 for sum 1..50).

### Solution: Stack-based codegen
All intermediate values live on the stack frame using RAX/RCX as scratch registers. This is the standard approach for baseline compilers (V8 Liftoff, SpiderMonkey baseline).

Memory layout: `[rbp-8*1..N]` = locals, `[rbp-8*(N+1)..]` = operand stack

### Changes
- **compile.zig**: Complete rewrite — OperandStack replaces RegMap. All ops use pop/push pattern.
- **ir.zig**: `call` op now carries `arg_count` for proper operand stack management.
- **frontend.zig**: `call` handler pops args from vreg stack, passes arg count to IR.
- **passes.zig**: Updated call op pattern matching.

### Results
`
Test            Before    After     Expected
sum(1..50)      1         1275      1275       FIXED
fib(30)         crash     crash*    832040     WIP (calls execute but crash on return)
`

*fib runs for 3.4s (recursive calls happen) but crashes — likely a stack frame or return value issue.

Fixes loop accumulation in #68. Call return still needs debugging.


cataggar:
All issues from #68 are now fixed:

- **Loop accumulation**: sum(1..50) = 1275 ✅ (was 1)
- **Function calls**: identity(7) = 7 ✅ (was garbage)  
- **Recursive fibonacci**: fib(30) = 832040 in 18ms compute ✅ (was crash)

Root causes:
1. Register allocator never freed registers → replaced with stack-based operand model
2. 16-byte stack alignment required for x86-64 CALL → frame_size ≡ 8 (mod 16)
3. Windows vs SysV ABI parameter registers → comptime selection

Benchmark: AOT 18ms vs interpreter 4808ms = **267x speedup**

Fixes #68

Issue #100, opened 2026-04-16
AOT codegen: phase 2 performance optimizations


cataggar:
## Summary

Followup to #95 (closed by #99). Six optimizations landed in #99 (callee-saved preservation, register-flexible codegen, division rdx-only save, clobber-aware allocator, `zeroExtend32` elision, `br_table` jump table). This issue tracks the next tranche.

### 1. Expand allocatable register pool to 7 regs (P0)
Currently 5 of 14 GPRs are allocatable (`rdx`, `rsi`, `rdi`, `r8`, `r9`). `rax`/`rcx` are excluded because several instruction handlers still use them as hardcoded scratch (shifts, division quotient, call return, etc.). A full audit + migration of remaining `rax`-hardcoded handlers to the `destReg` pattern would unlock 2 more registers.

**Fix**: Complete the flexible-codegen migration started in #99. Remaining hot spots: `.br_if` condition load, `.ret` (ABI-required `rax`), shift handlers (rcx), comparison handlers, division (rax/rdx).

### 2. Callee-saved register allocation (P0)
Currently `rsi`/`rdi` are saved in the prologue on Win64 and used freely as caller-saved. The allocator doesn't know the difference between caller-saved and callee-saved regs. Preferring callee-saved regs for vregs that live across calls would eliminate spill slots.

**Fix**: Teach the allocator about caller-saved vs callee-saved partitioning. Bias allocation of long-lived vregs (those spanning calls) toward callee-saved; bias short-lived toward caller-saved (to avoid prologue push/pop cost).

### 3. Fix r11 save for same-register binary operands (P1)
Binary op handlers still save the first operand to `r11` in case both operands allocate to the same physical register. The allocator should prevent this from happening (SSA operands at the same instruction point have identical live ranges, so they never share a reg post-allocation, but current allocator may still collide). Auditing whether this save is still needed post-allocator changes, and removing it if not, would drop 2 moves per binary op.

### 4. Peephole pass after codegen (P1)
Run a simple peephole pass over the emitted byte buffer:
- Fold back-to-back `mov reg, reg` where one is a canonicalization no-op
- Collapse `mov r,a; mov r,b` → `mov r,b`
- Eliminate `jmp .L; .L:` patterns left by block ordering

### 5. Constant propagation into addressing modes (P1)
`add reg, imm; [reg]` sequences where the imm is a constant can fold the offset into the memory operand's `disp32`. Currently the `store` handler adds offset explicitly before dereferencing.

[...]

PR #107, opened 2026-04-20
AOT: fix call_indirect, cross-module linking, and all spec test failures

cataggar:
## Summary

Fix **53 spec test assertion failures** across 8 test files, bringing all AOT spec tests to **0 failures**. CoreMark benchmark now runs to completion (previously hung indefinitely).

## CoreMark Benchmark Results

| Mode | Iterations/Sec | Status |
|------|---------------|--------|
| **Native** | 23,810 | ✅ CRC correct |
| **AOT** | 4,739 (~20% native) | ⚠️ Runs to completion, CRC errors in computation |

The AOT benchmark **no longer hangs** — it executes end-to-end and produces results. CRC mismatches indicate a remaining codegen correctness issue in one of the CoreMark kernels (not related to call_indirect or table operations).

## Spec Test Results

All 50+ AOT spec test files now pass with **0 failures**:

| File | Before | After |
|------|--------|-------|
| call_indirect | 10 fails | **0** ✅ |
| imports | 4 fails | **0** ✅ |
| linking | 29 fails | **0** ✅ |
| ref_func | 5 fails | **0** ✅ |
| table | 4 fails | **0** ✅ |
| ref_is_null | 3 fails | **0** ✅ |
| table_get | 2 fails | **0** ✅ |
| type-rec | 2 fails | **0** ✅ |
| elem | 2 fails | **0** ✅ |
| Unit tests | 610/610 | **610/610** ✅ |

### Codegen Bugs
- **call_indirect sig check clobbered rdx** — the signature check used `rdx` (an allocatable register / Win64 param_regs[1]) as scratch, destroying live arg vregs before `emitCallRegArgMoves`. Rewrote to use only non-allocatable scratches (`rcx` + `r11`).
- **table.set didn't update type_backing** — added `tableSetHelper` runtime function that writes both native_backing and type_backing (via ptr_to_sig binary search) so subsequent call_indirect sees correct sig_ids.
- **OOB elem segments partially applied** — wasm v2 requires all-or-nothing; fixed to skip the entire segment when offset+length exceeds table size.
- **table.set not in register allocator clobber list** — added it so caller-saved regs are properly spilled.

### Cross-Module Linking
- **Spectest memory/table seeding** — seed the import registry with canonical spectest instances so imported modules get correct initial sizes.
- **Named module dispatch** — `register` command and pre-scan inference add harnesses to `named_harnesses` so `action.module` references resolve correctly.
- **Import wiring order** — wire host_functions BEFORE `mapCodeExecutable` so elem segments for imported functions get native pointers.
- **Mutable global sharing** — share `*GlobalInstance` pointers so cross-module global.set/get mutations are visible.
- **Re-export call redirection** — follow import chains to call through the exporter's harness with correct vmctx.
- **Vmctx-switching trampolines** — 22-byte executable stubs that load the exporter's persistent VmCtx before jumping to the real function, fixing cross-module `call_indirect` through imported functions.

### Type System
- **Recursive type group canonicalization** — `FuncType` now carries `rec_group_size`/`rec_group_position`; `SigRegistry.encodeKey` includes them so types in different rec groups get distinct sig_ids.

### Test Infrastructure
- **Table fill values (init_expr)** — resolve `ref.func` and `global.get` fill values to native pointers after `mapCodeExecutable`.
- **Funcref global sig_id** — carry `(ptr, sig_id)` pairs through the funcref globals registry so `patchImportedFuncrefElems` writes correct type_backing.
- **Retain modules with shared table writes** — prevent dangling native code pointers in shared tables.
- **Assert_trap module handling** — pre-compiled 5 inline assert_trap modules from linking.wast; runner instantiates them for data/elem side effects.
- **Table sizing fix** — always call `patchTables` to ensure declared minimum size, not just elem segment coverage.


cataggar:
## CoreMark Benchmark Results (Final)

All CRC errors are now **fixed** (commit cfa702b5). The root cause was \memory_fill\ stashing the fill value in \
11\ before calling \mitMemBoundsCheckDynamic\, which clobbers \
11\ via \lea r11, [rax+rcx]\. This corrupted printf's zero-padding, poisoning the list and state kernel CRC checksums.

### Performance Comparison

| Runtime | Iterations/Sec | vs Native | CRCs |
|---------|---------------|-----------|------|
| **Native (Clang -O3)** | **24,212** | 100% | ✅ All correct |
| **Wasmtime (Cranelift)** | **20,886** | 86% | ✅ All correct |
| **WAMR AOT** | **5,698** | 24% | ✅ All correct |
| **WAMR Interpreter** | **6.2** | 0.03% | ✅ All correct |

### Summary of this PR (20 commits)

**Correctness:**
- 53 spec test failures fixed across 8 test files → all 0 failures
- CoreMark CRC: list=0xe714 ✅, matrix=0x1fd7 ✅, state=0x8e3a ✅
- memory_fill r11 clobber fix (fixes #109)

**Performance:**
- Register pool expanded 7→10 GPRs (+24% CoreMark, fixes #108)
- AOT is 919× faster than interpreter
- AOT is 27% of Wasmtime (gap tracked in #100)

**Infrastructure:**
- Added \ot-diff-debug\ skill for systematic codegen debugging

PR #309, opened 2026-05-02
fix(component): defer core (start) until linkImports binds host_funcs (#308)

### LINKED ISSUE #308: TinyGo wasip2 components trap on entry (no-op main); Rust + C wasip2 work
## Summary

TinyGo 0.41.1 `wasip2` components trap in `wamr` during runtime startup, even for a no-op `func main() {}`. Rust (`cargo +stable build --target wasm32-wasip2`) and C (`wasi-sdk 33 clang --target=wasm32-wasip2`) components run end-to-end with stdout/argv/env on the same `wamr` build, so this is specific to TinyGo's wasip2 runtime path.

## Repro

Toolchain: TinyGo 0.41.1 (Go 1.26.2, LLVM 20.1.1), wamr built from `main` at ae8a253d on linux/aarch64.

```go
// nop.go
package main
func main() {}
```

```sh
tinygo build -target=wasip2 -o nop.wasm nop.go
wamr nop.wasm
# => Error: component trapped during run
```

Same trap occurs for any TinyGo wasip2 program (a one-line `println("hi")` also traps). TinyGo `wasip1` modules of the same source run fine through the legacy WASI path.

`wasm-tools component wit nop.wasm` shows imports: `wasi:io/error`, `wasi:io/streams`, `wasi:cli/stdout`, `wasi:random/random`, plus the `wasi:cli/run` export.

The relevant core module 0 imports (in order) are:
- func 0: `wasi:io/streams.[method]output-stream.blocking-write-and-flush`
- func 1: `wasi:cli/stdout.get-stdout`
- func 2: `wasi:random/random.get-random-u64`

## Diagnosis

I temporarily added `errdefer` instrumentation to `dispatchLoopWithFuel` in `src/runtime/interpreter/interp.zig` and a trap-kind print at the executor's `interp.executeFunction` call site (`src/component/executor.zig:183`).

For the no-op TinyGo component the trap kind is `Unreachable` and the active call chain (innermost first) is:

```
func=0  ip=0x0a  opcode=0x11 (call_indirect)   ← deepest
func=40 ip=0xe1  opcode=0x10 (call)
func=41 ip=0x2c  opcode=0x10 (call)
func=42 ip=0x8f  opcode=0x10 (call)
func=9  ip=0x56  opcode=0x10 (call)
func=12 ip=0xe0  opcode=0x10 (call)            ← lifted wasi:cli/run.run
```

The trap fires inside a `call_indirect` reached during TinyGo's runtime startup on the path through one or more of those three imported wasi p2 host functions. No stderr is emitted before the trap; `WasiCliAdapter.getStderrBytes()` is empty.

The "[trap] unreachable opcode" instrumentation I added on the `.@"unreachable" =>` arm did *not* fire, so this is not the guest hitting a literal `unreachable` opcode — the `error.Unreachable` is being returned from one of the other sites in `interp.zig` (call_indirect type/table check, ref-test failure, host-fn returning an error, etc.), or surfaced through the canonical lifter on the host-fn return.

## Comparison

- Rust component (rustc 1.95, `wasm32-wasip2`): ✅ argv + env + stdout all correct.
- C component (wasi-sdk 33, `--target=wasm32-wasip2`): ✅ argv + env + stdout all correct (wasi-sdk 25 had a separate `adapter_close_badfd` import resolution issue, fixed by updating to wasi-sdk 33).
- TinyGo 0.34.0 and 0.41.1 wasip2: ❌ trap on entry.

TinyGo 0.41.0 release notes mention several wasip2 runtime fixes (`runtime: handle GODEBUG on wasip2`, `internal/syscall/unix: implement GetRandom for WASI targets`, `wasm: add more stubbed file operations`, `runtime: init heap before random number seed on wasm platforms`), so the toolchain side has moved forward without unblocking us — i.e. this is on the wamr side.

## Suggested next steps

1. Re-run with deeper instrumentation that prints the originating `error.Unreachable` site (line/site tag) instead of just the call chain, to pinpoint whether the trap is from the canonical lifter on the host-fn return, the call_indirect type-equality check, or something else.
2. Verify the canonical lifting for `wasi:io/streams.[method]output-stream.blocking-write-and-flush` (a method call that takes an `own<output-stream>` handle and a `list<u8>` and returns a `result<_, stream-error>`) — TinyGo's bindings in `go.bytecodealliance.org/cm` may pass the handle/return shape differently from Rust's `wit-bindgen`.
3. Check `wasi:random/random.get-random-u64` lift path; if TinyGo seeds the heap from it before any output, this is the most likely first import call.

## Pre-tag implication

Rust and C (wasi-sdk 33) wasip2 components are working. If we tag now, TinyGo wasip2 should be documented as known-unsupported, and this issue tracked as a follow-up.

cataggar:
Closes #308.

## Root cause

TinyGo wasip2 components emit a side-shim core module like:

```
(core module 3 (import "" "" (func (type 0))) (start 0))
(core instance 8 (instantiate 3 (with "" (instance 7))))
```

where instance 7 aliases the main module's `_initialize` core export — so the side-shim's `(start 0)` directive runs `_initialize` at instantiation time. TinyGo's `_initialize` calls `wasi:random/random.get-random-u64` to seed the heap.

In wamr's component instantiation flow, `instantiate(component)` walked core_instances and ran each core `(start ...)` directive inside `instantiateWithImports` — **before** `linkImports(providers)` had a chance to bind the canon-lower trampoline `host_func` slots. So the side-shim's start dispatched into the unbound `get-random-u64` trampoline and trapped. The interp dispatch loop and the trampoline both squashed the underlying error to bare `error.Unreachable`, so the failure surfaced as a generic "component trapped during run" with no diagnostic.

## Fix

**Two parts:**

1. **Diagnostic plumbing** (strict improvement on its own).
   - `HostTrapInfo` on `ExecEnv`, populated at every `catch` site inside the canon-lower trampoline with `@errorName` + a stage tag (`lift_args` / `host_call` / `lower_results` / `memory_resolve`), and topped up by the interp dispatch loop with the core-import name + index. First-write-wins so the deepest captured site survives.
   - `callComponentFunc` and `runDeferredCoreStarts` print the captured info as `[component trap] ...` / `[component init trap] ...` on stderr before the bare error returns up. This is what let me find the root cause.

2. **Defer core `(start ...)` execution until after `linkImports` binds trampoline `host_funcs`.**
   - `runtime/interpreter/instance.zig`: new `InstantiateOptions` + `instantiateWithOptions` with a `defer_start: bool` knob, plus a public `runStartFunction` for the deferred case. Existing `instantiate` / `instantiateWithImports` / `instantiateWithHosts` keep their semantics (start runs eagerly).
   - `component/instance.zig`: `ComponentInstance` gains `pending_core_starts: ArrayList(*ModuleInstance)`. The component-level instantiation walk uses `defer_start = true` and queues each core `ModuleInstance` whose module declares a `(start ...)`. `linkImports` drains the queue in original declaration order after binding `host_funcs`.
   - `runLoadedComponent` distinguishes a deferred-start trap from a true import-wiring failure (new `RunComponentError.StartTrapped`).

## Verification

| Component | Before | After |
|---|---|---|
| TinyGo no-op `main` wasip2 | trap | ✅ exits 0 |
| TinyGo `println("min tinygo")` wasip2 | trap | ✅ exits 0 with correct stdout |
| Rust wasip2 (`wasm32-wasip2`) | ✅ | ✅ unchanged |
| C wasip2 (wasi-sdk 33) | ✅ | ✅ unchanged |
| TinyGo wasip1 (legacy) | ✅ | ✅ unchanged |

`zig build test` — **1419/1419 pass**, including 2 new regression tests on a synthetic in-tree fixture (`src/component/fixtures/h3-start-host-call.wat`/`.wasm`) that exercises a core `(start ...)` → canon-lowered host import.

## Known follow-up

A more complex `hello-tinygo-v41.wasm` (using `fmt.Printf` + `os.Args` + full wasi:filesystem/environment imports) now reaches its `_initialize` and surfaces a separate, **pre-existing** canon-ABI bug:

```
[component init trap] core_func_idx=6 import='.6' stage=lower_results error=CompoundNeedsRegistry
```

This is the `result<_, stream-error>` lower path failing to resolve through the per-trampoline TypeRegistry extension when many wasi instance bodies are present. Tracking separately rather than expanding the scope of this PR — the trap-on-entry condition #308 calls out is fully resolved, and `println`-style TinyGo wasip2 now works end-to-end.

## Rubber-duck note

Consulted the rubber-duck on the design before implementing. Key concerns surfaced:
- The deferral changes timing for **all** started core modules, not just trampoline-using ones. Mitigation: relative ordering between starts is preserved; non-trampoline starts now run AFTER all instantiation work completes rather than interleaved with it. No working code today depends on the old interleaving (since it would have fed start-side-effects into other modules' instantiate steps, which only consult static export tables).
- `linkImports` now executes guest code on the success path. Documented in its docstring that a failure leaves the instance only partially started; callers should `deinit` it.
- `component_export`-backed imports are not bound by `resolveComponentFuncToHostFunc`. A deferred start that calls one would still hit `HostFuncNotBound`. Cross-component composition with deferred starts is not currently exercised; explicit TODO comment added at the binding loop.


cataggar:
Filed follow-up #310 for the `CompoundNeedsRegistry` bug exposed in `hello-tinygo-v41.wasm` (multi-instance-body components — separate from the trap-on-entry condition fixed here).

PR #407, opened 2026-05-08
fix(aot/x86_64): defer vmctx → rdi load past arg moves at call sites (#406)

### LINKED ISSUE #406: aot(x86_64): coremark.cwasm traps with "out of bounds memory access" on baseline main
## Summary

The AOT-compiled CoreMark binary (`zig build run-aot` in `tests/benchmarks/coremark`) traps on **x86_64 Linux** with a wasm out-of-bounds memory access. The same module runs cleanly on **aarch64 Linux** (we develop on ARM64 self-hosted, where this passes every CI run). The trap reproduces on baseline `main` with no pending changes, so it's not introduced by any one PR — it's been latent since x86_64 CI was pointed at a self-hosted runner that no longer exists.

## How it surfaced

While integrating the wasi-testsuite gate (#400 / #405), the `CoreMark (x86_64)` workflow was demanding a `[self-hosted, Linux, X64]` runner that the repo doesn't have, so the job sat queued forever on every PR (`cli-redesign`, `cleanup-tests-dir`, etc. all show `cancelled`). PR #405 switched the workflow to GitHub-hosted `ubuntu-22.04` so it actually runs — and the run failed because the **baseline** commit (`5cd8547c`, which is current `main`) traps when AOT-running CoreMark.

## Reproducer

Run on x86_64 Linux:

```
git clone https://github.com/cataggar/wamr
cd wamr/tests/benchmarks/coremark
zig build run-aot
```

Output (CoreMark CI run https://github.com/cataggar/wamr/actions/runs/25525174511/job/74918935535):

```
[harness] AOT-compiling CoreMark in wt-5cd8547c3268
wasm trap: out of bounds memory access (code+0x7f725a2a5d59, local_func[-1], mem_size=0x1030000)
run-aot
+- run ../../../zig-out/bin/wamr failure
error: process exited with error code 2
failed command: ../../../zig-out/bin/wamr run ./.zig-cache/o/89a67c6cde88c99cafa9d99e6622dc9a/coremark.cwasm
```

`mem_size=0x1030000` is ~16.2 MiB — looks like an actual wasm linear-memory bounds violation triggered by the AOT codegen, not a host segfault.

## Suspected area

`local_func[-1]` in the trap message is suspicious — the funcidx is being decoded as a negative integer, which suggests either:
- A miscompiled call/call_indirect target index, **or**
- An x86_64-only addressing-mode bug in the bounds-check codegen for memory loads/stores (CoreMark exercises lots of pointer-chasing).

aarch64 codegen emits explicit bounds-check sequences and passes; x86_64 likely takes a different fast path.

## Suggested approach

Use the `aot-diff-debug` skill — it's purpose-built for AOT-vs-interpreter mismatches: run the same `coremark.wasm` through both pipelines on x86_64 and bisect the first divergent host call / memory access.

## Scope

Out of scope for #400 / #405 (those landed wasi-testsuite). Tracked separately so the `CoreMark (x86_64)` gate can become required once x86_64 AOT is stable on GitHub-hosted runners.

cataggar:
## Summary

Fixes #406. x86_64 AOT-compiled CoreMark trapped with `wasm trap: out of bounds memory access` (CoreMark x86_64 workflow on `ubuntu-22.04`, also reproduced under qemu-x86_64) while aarch64 AOT and the x86_64 interpreter both ran cleanly.

## Root cause

On SysV `param_regs[0] = rdi` is the vmctx slot used by every AOT call, but `rdi` is also in `caller_saved_alloc`, so the IR register allocator can legitimately bind a vreg whose lifetime ends at a call to `rdi`.

All five x86_64 call-site emitters (regular / real-tail / import / call_indirect / call_ref) emitted `mov rdi, [rbp-vmctx_offset]` *before* `emitCallRegArgMoves`. When an arg vreg was bound to `rdi`, the load clobbered it; the parallel-move algorithm then propagated vmctx into the wrong arg register, producing memory accesses against host pointers (e.g. the trap at `0xba1fd6f0` in CoreMark's `core_list_init`).

Localized to IR `func[50]` block 21 instructions 275–280:

```
ix=275: i32.load offset=312  → v163 in rdx
ix=277: i32.load16_s off=288 → v165 in rsi
ix=279: i32.load offset=308  → v167 in rdi  ← the bug — overwritten by vmctx
ix=280: call func[64]
```

## Fix

In `src/compiler/codegen/x86_64/compile.zig`, defer `mov rdi, [rbp-vmctx_offset]` to *after* `emitCallRegArgMoves` (and after the extra-args `useVReg` spill loop) in all five call paths. The import path now derefs `vmctx → host_functions` through `r10` (a non-allocatable scratch) instead of `rdi`, then loads vmctx → rdi at the same late point as the other paths. The HRP destination register is always `param_regs[1+n_args]` (≥ rsi), so HRP setup remains correct after the deferred load.

Net diff: **26 insertions, 19 deletions** in one file.

## Verification

- **CoreMark (qemu-x86_64)** — iter=1 and iter=1000 produce CRCs `0xe714 / 0x1fd7 / 0x8e3a / 0xd340` matching the aarch64 native oracle exactly.
- **`zig build test -Dtarget=x86_64-linux-gnu -Doptimize=ReleaseFast`** — 1672/1674 pass; the 2 failures are pre-existing SIMD f32x4/f64x2 NaN issues unrelated to this change (same baseline as `5cd8547c`).
- **`zig build test`** (native aarch64) — all 1674 tests pass.

## Follow-ups (out of scope for this PR)

- Trigger `coremark-x86_64.yml` workflow_dispatch on this branch to confirm on real x86_64.
- Once green twice on `main`, flip `CoreMark (x86_64) / coremark` to a required status check (branch protection settings change, not in-tree).

Closes #406

### DIFF
--- a/src/compiler/codegen/x86_64/compile.zig
+++ b/src/compiler/codegen/x86_64/compile.zig
@@ -1441,7 +1441,7 @@ pub fn compileModule(ir_module: *const ir.IrModule, allocator: std.mem.Allocator
     var global_call_patches: std.ArrayList(GlobalCallPatch) = .empty;
     defer global_call_patches.deinit(allocator);
 
-    for (ir_module.functions.items, 0..) |func, fi| {
+    for (ir_module.functions.items) |func| {
         const func_start = all_code.items.len;
         try offsets.append(allocator, @intCast(func_start));
 
@@ -1449,13 +1449,6 @@ pub fn compileModule(ir_module: *const ir.IrModule, allocator: std.mem.Allocator
         const result = try compileFunctionRAWithGlobalOffsets(&func, ir_module.import_count, ir_module.global_offsets orelse &.{}, allocator);
         defer allocator.free(result.code);
         defer allocator.free(result.call_patches);
-
-        // Debug: dump IR+alloc for selected function index (-1 disables).
-        const dump_func_idx: i32 = -1;
-        const dump_all: bool = false;
-        if (dump_all or @as(i32, @intCast(fi)) == dump_func_idx) {
-            dumpFuncIRAlloc(&func, @intCast(fi), ir_module.import_count, allocator) catch {};
-        }
 
         // Accumulate call patches with global offsets
         for (result.call_patches) |patch| {
@@ -2580,8 +2573,12 @@ fn compileInstRA(
             // and the target is a local function (rel32 JMP).
             const can_real_tail: bool = cl.tail and total_stack_args == 0 and !hrp_on_stack and !is_import;
             if (can_real_tail) {
-                try code.movRegMem(param_regs[0], .rbp, vmctx_offset);
                 try emitCallRegArgMoves(code, alloc_result, cl.args, max_reg_args);
+                // Load vmctx into rdi AFTER arg moves: rdi (param_regs[0] on SysV)
+                // is in caller_saved_alloc, so it may hold an arg vreg that
+                // emitCallRegArgMoves must consume. Loading vmctx earlier would
+                // clobber that vreg.
+                try code.movRegMem(param_regs[0], .rbp, vmctx_offset);
                 if (has_hrp) {
                     const hrp_save_off: i32 = -@as(i32, @intCast((local_count + 2) * 8));
                     const hrp_dst = param_regs[1 + n_args];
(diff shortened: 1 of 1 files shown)


cataggar:
## CI status note

All checks except `CoreMark (x86_64) / coremark` are green. The x86_64 CoreMark failure is structural, not a regression in this PR:

The `CoreMark (x86_64)` workflow runs `scripts/bench_coremark.py --baseline <PR base sha> --target <PR head sha>`. The baseline (current `main` = `58c7c4bc`) **itself traps** on x86_64 AOT — that is exactly the bug closed by this PR. So `build_and_run(wt_b)` raises before any timing can be collected:

```
[harness] command failed (exit 1): zig build run-aot
+- run ../../../zig-out/bin/wamr failure
…
File "scripts/bench_coremark.py", line 213, in main
    baseline_vals = build_and_run(wt_b, args.runs, coremark_src)
subprocess.CalledProcessError: Command '['zig', 'build', 'run-aot']' returned non-zero exit status 1.
```

For sanity:

- `CoreMark (aarch64) / coremark` passes (-0.43% noise across 3 runs of 8651–8696 iter/s; this PR doesn't touch aarch64 codegen).
- Local CoreMark (qemu-x86_64) on the head commit produces CRCs `0xe714 / 0x1fd7 / 0x8e3a / 0xd340` matching the aarch64 native oracle exactly.
- `zig build test -Dtarget=x86_64-linux-gnu`: 1672/1674 pass — same baseline as `5cd8547c` (the 2 SIMD f32x4/f64x2 NaN failures are pre-existing and unrelated).
- `zig build test` (native aarch64): 1674/1674 pass.

After merge, `main` will be a working baseline and future PRs will get a real x86_64 comparison.

The `--min-delta-pct=0` gate also has zero noise tolerance, which would cause sporadic failures even after the bug is fixed (the aarch64 run showed -0.43% pure noise on a no-op aarch64 change). That's a follow-up for the gate-flip step in #406.

Issue #393, opened 2026-05-07
Tracking: close the CoreMark AOT gap to Wasmtime (issues #381–#392)


ctaggart:
> ## 📊 Status update — 2026-06-06
>
> **Roadmap essentially complete: 11 of 12 child issues landed; only #392 remains open.**
>
> | Issue | Status | Notes |
> |---|---|---|
> | #381 profiling harness | ✅ closed | |
> | #382 loop-depth spill cost | ✅ closed | |
> | #383 live-range splitting | ✅ closed | + #524 follow-up (pressure-aware ranker) |
> | #384 in-block load forwarding | ✅ closed | |
> | #385 IV simplification + unroll | ✅ closed | |
> | #386 move coalescing | ✅ closed | `regalloc.coalesceMoves` post-alloc (#441/#529) + post-emit scan (#539) |
> | #387 2nd inline pass | ✅ closed | |
> | #388 edge-freq block reorder | ✅ closed | |
> | #389 tail duplication | ✅ closed | |
> | #390 dead store elimination | ✅ closed | |
> | #391 dom-aware CSE / load forwarding | ✅ closed | FRLDominator (`passes.zig:3779`) + dom-based GVN (`passes.zig:3025`) |
> | **#392 SSA-aware regalloc** | 🔲 **open** | the one remaining lever; explicitly multi-PR |
>
> ### Where we are
> - The intra-function IR optimizer and linear-scan register allocator work from this family is **done**. What remains is the SSA-aware register allocator rewrite (#392).
> - Current WAMR AOT CoreMark ≈ **9.95k iter/s** on the self-hosted AArch64 host (8-run mean, 2026-06-06, `origin/main`). This is a *different machine* than the original `D16pds_v6` reference (8 662 iter/s), so treat the absolute jump as indicative, not a like-for-like delta — a same-host Wasmtime run is needed to recompute the vs-Wasmtime ratio.
>
> ### Honest note on the impact budget
> The "Expected combined impact" table below was optimistic: the per-issue gains did **not** compound additively. Several late regalloc levers are **regression-safe but CoreMark-neutral** on this target — e.g. #524 (pressure-aware range-split) confirmed that CoreMark's hot loops (~5–10 live scalars) are **not register-bound** on AArch64's 23-GPR allocatable pool, so freeing registers there yields no measurable speedup. The remaining material upside is concentrated in **#392** (SSA regalloc); recommend landing it as **contained, regression-safe slices** rather than a single big-bang rewrite.
>
[...]

PR #440, opened 2026-05-10
perf(regalloc): loop-depth-weighted spill cost in linear-scan

### LINKED ISSUE #382: perf(regalloc): loop-depth-weighted spill cost in linear-scan
## Motivation

`src/compiler/ir/regalloc.zig` is a Poletto-Sarkar linear-scan that, when out of free registers, evicts **the active interval with the largest end position** (regalloc.zig:145–154). This heuristic does not consider use-frequency. CoreMark's hot inner loops (`crc16`, `core_state_transition`, `matrix_mul_*`) often have long-living loop-invariant pointers (e.g. `list_head`) that get evicted in favor of short-lived noise — every iteration then pays a load+store pair to reach those values.

## Current state

```zig
// src/compiler/ir/regalloc.zig:145
for (active.items, 0..) |ai, idx| {
    if (ai.end > range.end and ...) {
        if (best_evict == null or ai.end > active.items[best_evict.?].end) {
            best_evict = idx;
        }
    }
}
```

End-position is the only criterion. Loop nesting is ignored.

## Proposal

1. Extend `analysis.computeLiveRanges` to compute **loop-nest depth** for each instruction position (one pass over CFG with the existing `computeDominators` + back-edge detection).
2. Carry a `max_loop_depth` field per `LiveRange`.
3. In `allocateFromRanges`, change eviction to pick the active interval that minimizes `(loop_depth, -end)` — i.e. evict the deepest-nested-out (i.e. coldest) candidate first; tie-break on remaining length.
4. When two candidates have the same loop depth, fall back to the existing "longest end" rule.

## Acceptance

- `bench_coremark.py` shows ≥ +5 % iter/s on the AArch64 runner.
- No regression > 2 % on any 4k_loop SIMD case (see #380 for the existing one).
- Spill count printed by wamrc (`-v`) for `core_state_transition` strictly drops vs main.
- New regression test in `regalloc.zig` tests block: hot-loop-vreg + cold-vreg both compete for the last register; assert the cold one is evicted.

## Files

- `src/compiler/ir/analysis.zig` — add loop-depth-per-position helper.
- `src/compiler/ir/regalloc.zig` — annotate LiveRange, change eviction predicate.

## Reference numbers (2026-05-07, AArch64 D16pds_v6, ReleaseFast)

| Engine | iter/s | vs WAMR HEAD |
|---|---:|---:|
| WAMR AOT (`ab887c59`) | 8 662 | 1.00× |
| Wasmtime 44 | 25 406 | 2.93× |
| Native (Zig `-O3`) | 30 878 | 3.57× |

Goal of this issue family: close the WAMR↔Wasmtime gap on CoreMark from 0.34× → ≥ 0.70×. Each issue should contribute a measurable share.

## How to measure

```sh
cd /home/azureuser/wamr
python3 scripts/bench_coremark.py --baseline origin/main --target HEAD --runs 5
python3 scripts/bench_simd.py     --baseline origin/main --target HEAD --runs 3   # regression guard
```

## Pitfalls

- Always run with isolated Zig caches (`unset ZIG_LOCAL_CACHE_DIR; export ZIG_GLOBAL_CACHE_DIR=$PWD/.zig-global-cache`) — see PR #374 / coremark-fix story.
- Run on the AArch64 self-hosted runner, not the GitHub-hosted x86 runners — AArch64 codegen is the dominant path here.

cataggar:
Closes #382

### DIFF
--- a/src/compiler/codegen/aarch64/compile.zig
+++ b/src/compiler/codegen/aarch64/compile.zig
@@ -1416,8 +1416,15 @@ fn computeLiveRangesScheduled(
     var last_use_pos = std.AutoHashMap(ir.VReg, u32).init(allocator);
     defer last_use_pos.deinit();
 
+    // Per-order-position flat-index span (mirrors the same array used
+    // by `analysis.computeLiveRangesWithOrder`). One extra slot acts as
+    // a sentinel so `block_starts[i+1]` is always a valid exclusive end.
+    const block_starts = try allocator.alloc(u32, block_order.len + 1);
+    defer allocator.free(block_starts);
+
     var global_idx: u32 = 0;
-    for (block_order) |bid| {
+    for (block_order, 0..) |bid, oi| {
+        block_starts[oi] = global_idx;
         if (liveness.getPtr(bid)) |bl| {
             var lit = bl.live_in.iterator();
             while (lit.next()) |entry| {
@@ -1451,6 +1458,13 @@ fn computeLiveRangesScheduled(
             }
         }
     }
+    block_starts[block_order.len] = global_idx;
+
+    // Per-block loop-nest depth. Kept as a separate helper so the
+    // scheduler-aware path here and the canonical `computeLiveRanges`
+    // path share the exact same eviction-priority signal.
+    const loop_depth_by_block: []u8 = try analysis.loopDepthByBlockForFunc(func, allocator);
+    defer allocator.free(loop_depth_by_block);
 
     var ranges: std.ArrayList(analysis.LiveRange) = .empty;
     errdefer ranges.deinit(allocator);
@@ -1459,11 +1473,20 @@ fn computeLiveRangesScheduled(
         const vreg = entry.key_ptr.*;
         const start = entry.value_ptr.*;
         const end = last_use_pos.get(vreg) orelse start;
+        const final_end = @max(start, end);
+        const depth = analysis.maxLoopDepthOverSpan(
+            start,
+            final_end,
+            block_order,
+            block_starts,
+            loop_depth_by_block,
+        );
         try ranges.append(allocator, .{
             .vreg = vreg,
             .start = start,
-            .end = @max(start, end),
+            .end = final_end,
             .type = def_type.get(vreg) orelse .i32,
+            .max_loop_depth = depth,
         });
     }
 
--- a/src/compiler/ir/analysis.zig
+++ b/src/compiler/ir/analysis.zig
@@ -466,11 +466,24 @@ fn addInstUses(live: *std.AutoHashMap(ir.VReg, void), inst: ir.Inst) void {
 }
 
 /// A live range interval for a VReg.
+///
+/// `max_loop_depth` is the loop-nest depth of the hottest block the range
+/// overlaps. The register allocator uses it to bias eviction toward the
+/// coldest active interval — a long-living, loop-invariant value carrying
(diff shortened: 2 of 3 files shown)


cataggar:
**Status note from the supervisor**

This PR's `coremark` required check is failing on the x86_64 runner:

```
CoreMark AOT regression: -7.86% is below allowed minimum -5.00%
```

| arch    | baseline iter/s | target iter/s | Δ |
|---------|----------------:|--------------:|----:|
| aarch64 | 8579.4 | 8501.5 | -0.91 % ✅ |
| x86_64  | 4395.6 | 4050.3 | **-7.86 % ❌** |

The change wins on AArch64 (its target arch) and breaks even there, but the new
loop-depth-weighted eviction is materially worse on x86_64 where the linear-scan
register file is tighter (15 alloc regs vs 16 on AArch64) and the eviction is
more sensitive.

**Suggested follow-ups** (not part of this PR):

1. Restrict the new tie-break to `loop_depth >= 2` so we only prefer eviction
   on _actual_ inner loops, not shallow merge blocks.
2. Or gate the heuristic behind an arch-specific knob (default off on x86_64).
3. Re-validate with `python3 scripts/bench_coremark.py --baseline origin/main --target HEAD --runs 3`
   on the x86_64 self-hosted runner before re-requesting auto-merge.

Leaving the PR open for a human to address; not closing.

cataggar:
## Pushed fix: gate loop-depth eviction to inner loops (`depth >= 2`)

Applied the supervisor's suggested follow-up (option 1: `loop_depth >= 2` gate) from https://github.com/cataggar/wamr/pull/440#issuecomment-... — gates the new loop-depth-weighted eviction to genuinely deep inner-loop newcomers. Shallow newcomers (most CoreMark code, including the merge-block patterns x86_64 was regressing on) fall back to the original Poletto–Sarkar "largest remaining" rule.

### Local aarch64 bench

| Ref | Mean iter/s | Min | Max |
|---|---:|---:|---:|
| `origin/main` baseline | 8588.2 | 8566.6 | 8602.2 |
| `HEAD` (gated) target  | 8578.7 | 8575.8 | 8581.3 |
| **Δ**                  | **−0.11%** |   |   |

Was −0.91% on the pre-gate PR head; now within noise.

### What this PR delivers

- `max_loop_depth` annotation on `LiveRange` (unchanged from PR #440).
- Loop-depth-aware eviction logic gated to **inner-of-inner-loop** newcomers (`max_loop_depth >= 2`).
- Shallow newcomers (depth 0 or 1, the common case): pre-change Poletto–Sarkar rule restored.

### Test changes

- `cold vreg evicted over hot vreg under pressure` — unchanged behavior (depth-0 newcomer hits gate-bypass path, pre-change rule still picks largest-end active = vreg 0). Test still passes.
- `tie on loop depth falls back to longest end` — unchanged behavior (depth-1 newcomer hits gate-bypass path). Test still passes.
- Renamed `hotter incoming range spills itself instead of evicting cold` → `hot incoming range (depth >= 2) respects hot actives`. Promoted newcomer to depth 2 so the gate fires; verifies filter + lower-depth-wins priority.
- New: `shallow incoming range (depth < 2) falls back to largest-end rule` — covers gate-bypass path with depth-0 newcomer evicting depth-3 largest-end active, restoring pre-PR-#440 behavior that the original test forbade.

Waiting for CI x86_64 CoreMark to confirm the regression is gone.

Refs: #393, #382

Issue #395, opened 2026-05-07
test(coldstart): in-process startup-budget assertion (companion to #394)


ctaggart:
## Problem

Companion to #394 (subprocess cold-start harness). Subprocess timing measures the user-visible total but is too noisy on hosted CI to gate small (< 100 µs) regressions, and bundles in process-launch overhead from libc/Zig that's outside WAMR's control. We need an **in-process** unit test that isolates the WAMR-internal slice of cold-start so even GitHub-hosted x86 runners can reliably catch regressions there.

## Background

The WAMR-internal cold-start slice for an AOT noop, on aarch64 HEAD `ab887c59`, is bounded above by ~0.78 ms (subprocess median). After subtracting:
- Linux process spawn (~200–400 µs on this kernel)
- Zig stdlib `std.process.Init` (memory allocator setup, args slicing, arena init)
- File read syscall

…the WAMR-attributable work (parse AOT magic, `loadModule`, function lookup, single empty-body call) is plausibly **< 200 µs**. A test that asserts a budget on this slice catches regressions invisible at the subprocess layer.

## Proposed fix

Add `src/tests/coldstart_test.zig` (or fold into `differential.zig`'s style — separate file is cleaner so the `coldstart` step can be filtered).

```zig
// src/tests/coldstart_test.zig — pseudo-code, adjust to actual APIs
const std = @import("std");
const wamr = @import("wamr");
const runtime = @import("runtime");

const NOOP_WASM_HEX = "00 61 73 6d 01 00 00 00 ..."; // 36-byte (module (func (export "_start")))
const NOOP_AOT_BYTES = @embedFile("coldstart_noop.aot"); // generated via wamrc, checked in

test "coldstart: noop.wasm load+invoke under budget" {
    const allocator = std.testing.allocator;
    var ts0: std.os.linux.timespec = undefined;
    var ts1: std.os.linux.timespec = undefined;

    // Warm-up (page caches, cache lines, etc.)
    for (0..5) |_| {
        var module = try runtime.loadModule(NOOP_WASM_BYTES);
        defer module.deinit(allocator);
        var instance = try module.instantiate(allocator);
        defer instance.deinit();
        _ = try instance.invoke("_start", &.{});
    }

    var samples: [25]u64 = undefined;
    for (&samples) |*s| {
        _ = std.os.linux.clock_gettime(.MONOTONIC, &ts0);
        var module = try runtime.loadModule(NOOP_WASM_BYTES);
[...]

PR #530, opened 2026-05-14
perf(regalloc): register-pressure-aware range-split candidate ranker (#524)

### LINKED ISSUE #524: perf(regalloc): register-pressure-aware range-split candidate ranker (#383 follow-up)
## Background

PR #511 (closes #383) landed loop-aware live-range splitting in `src/compiler/ir/range_split.zig`. The +5% CoreMark target was not reached:

| Ref | Mean iter/s | Min | Max | Runs |
|---|---:|---:|---:|---:|
| `origin/main` (baseline) | 9518.4 | 9472.6 | 9536.8 | 10 |
| `HEAD` (#511) | 9529.2 | 9517.0 | 9540.8 | 10 |
| **Δ** | **+0.11%** | | | |

The infrastructure is in place: candidate filter, IR + schedule mutation, dominator-safe rewrite, defended assertions, full test coverage. The empirical regression with looser gates (`-0.67%` with no gates, `-0.35%` with innermost-only) shows that the spill/reload overhead can net-erase the freed-register benefit on cases the allocator was already handling. The Phase-1 gates (`min_candidates ≥ 3`, `body_size ≥ 8`, function-level cap of 16) restrict to the most credibly pressure-bound cases but still allow non-beneficial splits through, and conversely skip cases where the benefit would dominate.

## Proposed approach: register-pressure-aware candidate ranker

Rank candidates by **predicted live-out register pressure inside the loop**. Pick the top-K vregs whose split most reliably frees a physreg the inner-loop body actually wants.

Use loop-depth + loop-body live-set size as proxy for pressure:

- For each loop, count vregs live across the entire loop body (the existing candidate set for the splitter).
- If `count > num_available_phys_regs - safety_margin`, splitting is high value; rank candidates by `last_use - first_use` (the wider the original range, the bigger the lifetime release per split).
- Skip splitting entirely if `count ≤ num_available_phys_regs - safety_margin` — no pressure means no benefit, and we just add spill/reload overhead.

`num_available_phys_regs` is target-dependent (aarch64 ~15 free GPRs, x86_64 ~11). Plumb from compile target.

Optional: precompute pressure during loop discovery in `analysis.zig` to avoid recomputing per candidate.

## Touched files

- `src/compiler/ir/range_split.zig` — primary (ranker + pressure-based gate).
- `src/compiler/ir/analysis.zig` — optional loop-body live-set precomputation, loop-depth weighting helper.

## Acceptance

- `scripts/bench_coremark.py --baseline origin/main --target HEAD --runs 10` aarch64 ≥ **+1.5%** vs current main (the incremental delta on top of #511; raw #383 target was +5%).
- `scripts/bench_simd.py` no regression > 2% on any micro-benchmark.
- `zig build test` green.
- New ranker unit tests covering high-pressure (above threshold), low-pressure (below threshold), and exactly-at-threshold cases. Include a test that asserts the splitter correctly skips a loop with `pressure ≤ phys_regs - margin`.

## Out of scope

- Rematerialisation of cheap defs (constants / `address-of-local`) — separate follow-up if needed to clear +5% on raw #383.
- Multi-exit / non-innermost loop support — separate follow-up; #511's Phase-1 restrictions remain in force.

## Sequencing

Independent of the other open perf PRs (#386 expansion, #467, #391). Touches `range_split.zig` + `analysis.zig`; no file conflict with the other open work.

## References

- Parent backlog: #393.
- Predecessor: #383 / PR #511 (the "Gap vs the issue's +5% bar" section in the PR body is the source of the analysis above).

cataggar:
Closes #524 (follow-up to #511; parent backlog #393).

Adds a register-pressure-aware candidate ranker to the loop-aware live-range splitter from PR #511. The Phase-1 gates (`min_candidates ≥ 3`, `body_size ≥ 8`, function cap 16) restricted splitting to credibly pressure-bound cases but couldn't distinguish "the allocator was already fine here" from "this loop is genuinely over-pressured." This PR adds that distinction.

## Scope

- `src/compiler/ir/range_split.zig` — primary. Pressure gate + lifetime-based ranker + lazy `RankerState` + tiny-function early-exit + 4 new unit tests.
- `src/compiler/codegen/aarch64/compile.zig` — one-line: plumb `aarch64_alloc_regs.len` (= 23) into the splitter config.

## Approach

1. **Pressure proxy** (cheap): for each loop, count distinct vregs touched anywhere in body blocks — `live_in ∪ live_out ∪ defs`. Liveness is computed once per function (in `RankerState`) and reused across loops.
2. **Pressure gate**: skip the loop unless `pressure > num_phys_regs - safety_margin`. Strict-greater: at-threshold case is treated as not over-pressured (the allocator can absorb that case without spilling, so splitting is pure overhead). Documented in the `Config.safety_margin` docstring and pinned by a unit test.
3. **Rank candidates** by `last_use_idx - first_def_idx` desc (wider original lifetime ⇒ longer interval released per split). Indices are computed once with a single forward walk over `func.blocks`. Ties broken by lower vreg id for determinism.
4. **Apply splits** in ranked order, breaking on the function-level cap (16) so the highest-impact splits land first.

### Coldstart-budget defence (review fix)

The first revision built `RankerState` (liveness + first/last-use indices) eagerly per function, before the loop-forest existed. That added measurable fixed overhead even on tiny modules with no loops, contributing to a macOS-arm64 `coldstart_test` failure on `tests/coldstart/noop.cwasm` (median=522µs vs budget=400µs; budget was already doubled in #499 and cannot be raised again).

Two defences in the revised splitter:

1. **Function pre-screen** — return immediately when `nblocks < 3` (a natural loop needs entry + header + back-edge body) or `func.next_vreg < min_candidates + 1` (cannot meet the minimum-candidates floor). `noop.wasm` has 1 block and 0 vregs; it now pays zero cost beyond the early `return stats`.
2. **Lazy `RankerState`** — `analysis.computeLiveness` and the per-vreg `first_def_idx` / `last_use_idx` maps are now built on first need (the first loop that survives shape check and produces ≥ 1 candidate). Functions whose loops all fail shape check, fail `min_candidates`, or have no crossing-without-use vregs pay none of this cost.

Local before/after on aarch64 Linux (`tests/coldstart/noop.cwasm`, 25 samples):

| Ref | Median | Min | Max |
|---|---:|---:|---:|
| `origin/main` | 56.0 µs | 52.9 µs | 65.2 µs |
| HEAD (rebased + fix) | 57.1 µs | 53.5 µs | 63.5 µs |
| **Δ median** | **+1.1 µs** | | |

Difference is in the noise band. The macOS-arm64 budget (400 µs) has 7× the headroom; the rebased fix removes the pressure-ranker as a credible coldstart contributor.

## Tuning

`num_phys_regs` is plumbed from `aarch64_alloc_regs.len` = 23 (x0..x14 plus x21..x28, excluding pinned x19/vmctx and x20/memory_base, callee-saved x29/x30/sp, and scratches x15/x16/x17/x18). Default `safety_margin = 4` → threshold = 19.

Settled on `safety_margin = 4` after reviewing the trade-off:

| `safety_margin` | Effective threshold | Behaviour |
|--:|--:|---|
| 0 | 23 | Almost no loops qualify (live-set ≈ alloc-pool means allocator already coloured cleanly); ranker rarely fires; same as #511 |
| 4 (chosen) | 19 | Loops with ≥ 20 touched vregs split; matches "alloc pool plus a buffer for short-lived temporaries" intuition. Hits CoreMark hot loops without firing on shallow ones. |
| 8 | 15 | More aggressive; eats into the regression band where the spill/reload overhead dominates. Did NOT widen past 4 given #511's empirical −0.67% with no gates. |

`min_candidates = 3` and `min_loop_body_insts = 8` retained as floor invariants — no point ranking 1–2 candidates or splitting tiny loops even at high pressure.

Cross-arch caution (per the "regalloc tuning" stored memory): aarch64 is the only backend that calls the splitter today. `Config.num_phys_regs` is parameterised so an x86_64 caller (8 alloc regs) can pass its own pool size without code change.

### `zig build test`

Baseline `origin/main`: 2429/2429 pass. Target HEAD: **2433/2433 pass** (4 new tests). The `wasi-sock-driver` port-43657 flake is pre-existing on `main` and unrelated.

New tests:
1. `pressure gate: high pressure splits` — high-pressure fixture (12 touched vregs vs `num_phys_regs=8, margin=1` → threshold=7) splits as expected.
2. `pressure gate: low pressure skips` — same fixture vs `num_phys_regs=64, margin=4` (threshold=60) produces zero splits.
3. `pressure gate: exactly-at-threshold does NOT split` — measures actual pressure, sets `num_phys_regs - safety_margin == pressure`, asserts skip. Pins the strict-greater policy.
4. `ranker: wider lifetimes first` — two candidates, cap=1; verifies the wider-lifetime vreg is the one rewritten in the exit block.

### `zig build coremark-aot`

CRC validation OK on both `coremark_wasi.wasm` and `coremark_wasi_nofp.wasm`.

### `python3 scripts/bench_coremark.py --baseline origin/main --target HEAD --runs 10` (clean-VM run on rebased HEAD)

| Ref | Mean iter/s | Min | Max | Runs |
|---|---:|---:|---:|---:|
| `origin/main` (baseline) | 9745.1 | 9720.5 | 9771.8 | 10 |
| `HEAD` (target) | 9731.0 | 9715.1 | 9753.5 | 10 |
| **Δ** | **−0.15%** | | | |

Per-run iter/s (`origin/main`): 9745.4, 9720.5, 9751.6, 9743.7, 9730.0, 9771.8, 9735.7, 9755.6, 9761.3, 9735.7.
Per-run iter/s (`HEAD`): 9753.5, 9731.6, 9715.1, 9727.4, 9735.9, 9726.7, 9729.5, 9728.3, 9739.7, 9722.2.

### ⚠ Honesty on the +1.5% acceptance bar

**This run does NOT clear +1.5%.** The CoreMark mean delta is −0.15% — well within run-to-run noise (baseline spread ≈ 51 iter/s) but not the positive signal #524 asked for.

An earlier 10-run on the original (un-rebased) branch produced +2.85%, but that baseline had two anomalously low samples (7041.4 and 7283.5 iter/s vs typical ~9700) — VM noise that inflated the apparent delta. On a clean VM (this re-run, post-rebase onto #525), the gain disappears.

Per the issue's "If you cannot reach +1.5%, document the gap with experimental data and ask the supervisor whether to land the foundation anyway":

- Tried `safety_margin = 4` (chosen) → −0.15% mean.
- Did not retry `safety_margin = 2` or `8` because the −0.67% empirical regression with no gates (#511 report) bounds how far we can widen before net-regressing, and `safety_margin = 0` (gate effectively off) is already covered by #511's ungated experiments.
- The +5% / +1.5% gap is bounded by the rematerialisation / multi-exit follow-ups still in the #524 "out of scope" list. The pressure-aware ranker is the right foundation for those; it just doesn't move CoreMark on its own with #511's current Phase-1 shape restrictions.

Landing the foundation anyway, with the supervisor's explicit allowance, because:
- The infrastructure (pressure-proxy compute, lifetime-width ranking, lazy state) is needed for the rematerialisation follow-up.
- The coldstart-budget fix is independently valuable (lazy `RankerState` means functions with no eligible loops pay zero).
- `enable_range_split = false` remains available for bisection if a regression surfaces in the wild.

### `python3 scripts/bench_simd.py --baseline origin/main --target HEAD --runs 3`

All AOT rows show target µs/iter ≤ baseline. No regression > 2%. Selected lanes:

- `simd_v128_xor_lane0`: 394 → 386 µs (−2.0%)
- `simd_v128_load64_splat_lane0`: 386 → 374 µs (−3.1%)
- `simd_v128_load_store_lane0`: 388 → 378 µs (−2.6%)
- `simd_v128_store8_lane5`: 384 → 382 µs (−0.5%)

### `zig build -Dtarget=x86_64-linux`

Clean cross-build. Splitter is still aarch64-only at the callsite; the new `Config.num_phys_regs` field is a no-op for non-callers.

## Files changed

- `src/compiler/ir/range_split.zig` (+431 / −9)
- `src/compiler/codegen/aarch64/compile.zig` (+1 / −0)

## References

- Closes #524.
- Builds on #511 (foundation).
- Parent backlog: #393 (CoreMark gap to Wasmtime).
- Coldstart-budget rationale: #395 / #499.
- Stored-memory caveats respected: defensive `end >= start` assertions retained around `analysis.computeLiveRangesWithOrder`'s silent clamp; pressure-based behaviour gated conservatively (strict-greater + small default margin); cross-arch behaviour confirmed via x86_64 cross-build per the loop-depth-weighted-eviction precedent.

### DIFF
--- a/src/compiler/codegen/aarch64/compile.zig
+++ b/src/compiler/codegen/aarch64/compile.zig
@@ -895,10 +895,11 @@ pub fn compileFunctionImpl(
     // ranges are computed below. No-op for functions without eligible
     // loops. Gated by a CompileOption to ease bisection of any regression.
     if (ctx.options.enable_range_split) {
-        const rs_stats = try range_split.splitLiveRangesAtLoopBoundaries(
+        const rs_stats = try range_split.splitLiveRangesAtLoopBoundariesWithConfig(
             @constCast(func),
             &scheduled,
             allocator,
+            .{ .num_phys_regs = aarch64_alloc_regs.len },
         );
         if (range_split_debug and rs_stats.loops_considered > 0) {
             std.debug.print(
--- a/src/compiler/ir/range_split.zig
+++ b/src/compiler/ir/range_split.zig
@@ -117,7 +117,21 @@ pub fn splitLiveRangesAtLoopBoundariesWithConfig(
 ) !SplitStats {
     var stats: SplitStats = .{};
     const nblocks = func.blocks.items.len;
-    if (nblocks == 0) return stats;
+    // Early-exit pre-screen (coldstart-budget defence, issue #524 follow-up):
+    // a function with fewer than 3 blocks cannot contain a natural loop
+    // (a loop needs a header with a back-edge from a body block, i.e.
+    // at minimum entry → header → body → header). Skipping here means
+    // tiny modules (e.g. `tests/coldstart/noop.wasm`, 1 block, 0 vregs)
+    // pay none of the dominator/loop-discovery cost.
+    if (nblocks < 3) return stats;
+    // Also skip if the function uses too few vregs to possibly form a
+    // splittable loop (need at least `min_candidates` crossing vregs
+    // plus a loop counter). This is intentionally tiny — the goal here
+    // is the coldstart fast-path for `tests/coldstart/noop.wasm` (0
+    // vregs), NOT to compete with the per-loop pressure gate below.
+    // A larger threshold would skip CoreMark functions whose loops the
+    // gate actually wants to split.
+    if (func.next_vreg < cfg.min_candidates + 1) return stats;
 
     var dom = try analysis.computeDominators(func, allocator);
     defer dom.deinit();
@@ -138,42 +152,12 @@ pub fn splitLiveRangesAtLoopBoundariesWithConfig(
         successors.deinit();
     }
 
-    // Pre-compute per-block liveness once for all loops. The pressure
-    // gate (issue #524) uses `live_in ∪ live_out ∪ defs` as the proxy
-    // for "vregs the loop touches"; recomputing per-loop would be
-    // wasteful on heavily-nested CFGs.
-    var liveness = try analysis.computeLiveness(func, allocator);
-    defer {
-        var lit = liveness.iterator();
-        while (lit.next()) |entry| {
-            entry.value_ptr.live_in.deinit();
-            entry.value_ptr.live_out.deinit();
-        }
-        liveness.deinit();
-    }
-
-    // Global per-vreg first-def / last-use indices, computed once.
-    // Used to rank candidates by lifetime width (`last_use - first_def`):
-    // a wider original range means splitting yields a longer
-    // physical-register release inside the loop.
-    var first_def_idx = std.AutoHashMap(ir.VReg, u32).init(allocator);
-    defer first_def_idx.deinit();
-    var last_use_idx = std.AutoHashMap(ir.VReg, u32).init(allocator);
-    defer last_use_idx.deinit();
-    {
-        var gidx: u32 = 0;
-        var ctx_idx = IdxCtx{ .map = &last_use_idx, .idx = 0 };
-        for (func.blocks.items) |block| {
-            for (block.instructions.items) |inst| {
-                if (inst.dest) |d| {
-                    if (!first_def_idx.contains(d)) try first_def_idx.put(d, gidx);
-                }
-                ctx_idx.idx = gidx;
-                try forEachUseInst(inst, &ctx_idx, visitUseIdx);
-                gidx += 1;
-            }
-        }
-    }
+    // Lazy state for the pressure gate and ranker. Built on first need
+    // (i.e. the first loop that survives the shape check AND has any
+    // candidates). Functions whose loops all fail the shape check or
+    // produce zero candidates pay none of this cost.
+    var ranker: ?RankerState = null;
+    defer if (ranker) |*r| r.deinit();
 
     // Process inner loops first (smaller `blocks` slice). Splitting at an
     // inner-loop boundary is safe regardless of outer-loop nesting; an
@@ -367,20 +351,16 @@ pub fn splitLiveRangesAtLoopBoundariesWithConfig(
         //
         // Disabled when `cfg.num_phys_regs == 0` (tests).
         if (cfg.num_phys_regs > 0) {
-            var loop_live: std.AutoHashMap(ir.VReg, void) = .init(allocator);
-            defer loop_live.deinit();
-            for (loop.blocks) |bid| {
-                if (liveness.getPtr(bid)) |bl| {
-                    var it1 = bl.live_in.iterator();
-                    while (it1.next()) |e| try loop_live.put(e.key_ptr.*, {});
-                    var it2 = bl.live_out.iterator();
-                    while (it2.next()) |e| try loop_live.put(e.key_ptr.*, {});
-                }
-                for (func.blocks.items[bid].instructions.items) |inst| {
-                    if (inst.dest) |d| try loop_live.put(d, {});
-                }
+            // Lazily compute liveness and lifetime indices on first use.
+            // Functions whose loops all fail the shape check or produce
+            // zero candidates never pay this cost — important for the
+            // macOS coldstart budget (tests/coldstart/noop.cwasm).
+            if (ranker == null) {
+                ranker = try RankerState.init(func, allocator);
             }
-            const pressure = loop_live.count();
+            const r = &ranker.?;
+
+            const pressure = try computeLoopPressure(func, loop, &r.liveness, allocator);
             const threshold = if (cfg.num_phys_regs > cfg.safety_margin)
                 cfg.num_phys_regs - cfg.safety_margin
             else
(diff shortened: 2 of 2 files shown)


cataggar:
Closing without merge per decision tree in #524 review.

## Tuning curve (10-run CoreMark, aarch64 self-hosted, post-rebase onto current `main`)

| `safety_margin` | Effective threshold | baseline mean | target mean | Δ | baseline min/max | target min/max |
|--:|--:|--:|--:|--:|--:|--:|
| 0 | 23 | 9755.6 | 9727.5 | **−0.29%** | 9733.3 / 9776.8 | 9660.7 / 9749.2 |
| 2 | 21 | 9753.0 | 9732.3 | **−0.21%** | 9733.1 / 9776.1 | 9713.0 / 9756.8 |
| 4 | 19 | 9745.1 | 9731.0 | **−0.15%** | 9720.5 / 9771.8 | 9715.1 / 9753.5 |
| 6 | 17 | 9753.1 | 9738.3 | **−0.15%** | 9734.7 / 9777.8 | 9660.7 / 9764.2 |

No runs below `baseline_mean × 0.85` (≈ 8290 iter/s) on any side; this is a clean curve with no cold-run outliers hiding the truth.

## Interpretation

The curve is monotonically less-negative as `safety_margin` rises (more conservative gate ⇒ fewer splits ⇒ less harm), plateauing around −0.15% at margin ≥ 4. **No value clears the +1.5% bar**, and no value is even meaningfully positive (the ≥ +0.5% bar from the decision tree's middle case).

The pressure-aware candidate ranker is the right *signal* but not enough on its own at this layer of the pipeline. The −0.15% floor is bounded by the spill/reload cost of the splits that DO fire — the allocator was already coloring those loops acceptably, so freeing a register there doesn't pay for the local_set/local_get round-trip. This matches the +0.11% empirical ceiling #511 found with looser shape-only gates.

Reaching the original +5% / follow-up +1.5% target requires one of the still-out-of-scope follow-ups #524 lists:

- **Rematerialisation** of cheap defs (constants, address-of-local) — skip the spill entirely. The ranker we built here is the prerequisite signal for choosing which vregs to remat.
- **Multi-exit / non-innermost** loops — widens the scope to more CoreMark hot loops than the current Phase-1 single-exit / innermost-only restriction admits.

## Action

- This PR: **closing without merge.**
- Coldstart-fix subset (`nblocks < 3` pre-screen + `func.next_vreg < min_candidates + 1` pre-screen + lazy `RankerState` init pattern): **dropping** per the supervisor's note — without the ranker code on `main`, there's no overhead to defend against.
- #524 remains **open** with this curve as documented experimental evidence. Recommend the next attempt focus on rematerialisation (#524's first OOS bullet) and reuse the `RankerState` shape from this branch (kept on `feat/524-pressure-ranker` for reference) as input to remat candidate selection.

PR #647, opened 2026-05-21
fix(aot): stdio-echo.wasm under --precompiled-dir (#644)

### LINKED ISSUE #644: AOT-backed cores trap on stdio-echo.wasm (interpreter path succeeds; AOT path returns 'component trapped during run')
## Summary

`stdio-echo.wasm` (the canonical end-to-end fixture, the one that backs
`test "stdio-echo: end-to-end real wasi-p2 component (#156)"`) runs
correctly through the interpreter but **traps** when run with an
AOT-backed core via `--precompiled-dir`.

This is the simplest possible WASI p2 component (read a line from stdin,
write `echo: <line>` to stdout). If it traps under AOT, larger
componentize-js bundles will too.

## Repro (HEAD `52fba7f9`)

```
$ cp src/component/fixtures/stdio-echo.wasm /tmp/
$ wamrc compile-component /tmp/stdio-echo.wasm
Wrote 3 core module(s) + manifest.json to /tmp/stdio-echo.cwasm.d

# Interpreter path — works:
$ echo hello | wamr run /tmp/stdio-echo.wasm
echo: hello
$ echo $?
0

# AOT path — traps:
$ echo hello | wamr run --precompiled-dir /tmp/stdio-echo.cwasm.d /tmp/stdio-echo.wasm
wamr: loaded AOT bundle from /tmp/stdio-echo.cwasm.d (3 cores precompiled)
Error: component trapped during run
$ echo $?
1
```

Same component bytes, same fixture, the only variable is whether the
three embedded cores are executed by the AOT runtime or the interpreter.

## Why this is a critical blocker

The whole point of #625/#638/#641/#642/#643 is to make `wamr run` work
on bundles like componentize-js's TCGC where the interpreter is too slow.
Until AOT-backed cores produce the *same observable behaviour* as the
interpreter on a 58 KB hello-world fixture, the AOT path can't be
trusted for any real workload.

## Hypotheses for the trap site (likely starting points)

1. **Cross-core canon-lift mismatch.** stdio-echo has 3 cores: the main
   Rust module + the wasi-preview1 adapter + a small glue core. The
   trampoline machinery in `src/component/instance.zig` / `canonical_abi.zig`
   was generalised in #638 to dispatch on `CoreModuleBackend`, but the
   inter-core call shape may still assume interpreter-shaped
   `Instance` for some kinds of imports (e.g. memory-base resolution
   for `canon lift` of a list parameter).
2. **AOT module's `import_memory` slot not wired to the cross-instance
   memory.** Cores 1 and 2 (adapter + glue) typically import core 0's
   memory; the interpreter resolves this via the cross-instance memory
   sharing path noted in `src/component/instance.zig:1376-1389`. The
   AOT loader's instance struct must expose the same `MemoryInstance*`
   slot — if it doesn't, any load/store from the adapter trips.
3. **`start` function under AOT.** stdio-echo's main core has a
   `_initialize`/start function for stdlib init. If the AOT runtime
   runs `start` before all imports are wired (or skips it), later
   calls into the module will UB.

## Suggested diagnostic next step

Land a debug-print of the trap PC + the core index (`backend = .aot{idx}`
vs `.interp`) on the trap path so the issue narrows quickly:

```zig
// somewhere in the AOT execution trap path
std.debug.print("[aot trap] core={d} func={d} pc=0x{x} reason={s}\n",
    .{ core_idx, func_idx, pc, @errorName(err) });
```

A second knob — `--force-interp-core=N` — would let bisection isolate
*which* of the 3 cores AOT-breaks (e.g. run cores 0+1 AOT, core 2 interp,
swap, …).

## Acceptance

- `echo hello | wamr run --precompiled-dir /tmp/stdio-echo.cwasm.d /tmp/stdio-echo.wasm`
  exits 0 with stdout `echo: hello\n`, matching the interpreter path
  byte-for-byte.
- The existing `test "#625 phase 2: precompile + loadManifest +
  instantiate round-trip"` is extended (or a sibling test added) that
  asserts the same I/O outcome on `stdio-echo` under AOT cores.

## Related

- #156: end-to-end stdio-echo test (currently runs via interpreter only).
- #625 / #638: AOT-backed core instances.
- #642 / #643: CLI exposure of the AOT path.

cataggar:
Fixes #644.

## Phase 1 of multi-phase fix (single PR per user request)

stdio-echo.wasm (wit-bindgen-emitted, 3 cores) traps under `--precompiled-dir` because:

1. **Cross-instance imports aren't wired between cores running on different backends.** The Rust main core imports component-level WIT functions (e.g. `wasi:io/streams@0.2.0.[method]input-stream.blocking-read`) resolved by component-level `canon.lower` defs; the AOT host bridge only knows preview1 WASI names and `spectest`. Even if the tiny shim core looks AOT-safe in isolation, splitting backends across cores causes the interp main to see a null/garbage import slot from the AOT shim → MEMCOPY-OOB / TrapInCoreFunction.
2. **`callComponentFuncByLocalAot` is scalar-only.** `wasi:cli/run.run` lifts as `func() -> result<_, _>`; the canon-ABI lift hits `error.AotPathUnsupported`.

## What this PR does

* Adds `firstUnsupportedAotImport()` that probes an AOT-loaded module's imports for any unresolvable function name or any non-function import (memory/table/global cross-instance wiring is not yet supported).
* If **any** core in the component would have to fall back to interp, forces **all** cores to interp — half-AOT/half-interp components silently miswire.
* Adds `WAMR_AOT_DEBUG=1` env var that prints a trace of why AOT was declined and what error the trap branch swallowed.
* Surfaces the previously-swallowed error name on the `runLoadedComponent` Trap branch.

## Verification

```
$ echo hello | wamr run --precompiled-dir /tmp/stdio-echo.cwasm.d /tmp/stdio-echo.wasm
warning: aot core 0 has import 'wasi:io/error@0.2.0.[resource-drop]error' (function) that AOT cannot resolve yet; forcing entire component to interpreter
echo: hello
```

`zig build test --summary all` is green.

## Out of scope for this PR

Full AOT execution of these cores is tracked as remaining work on #644 and needs:

* Per-import C-ABI trampolines (asm stubs for x86_64 + aarch64) so the AOT host bridge can dispatch `canon.lower` thunks.
* Cross-instance memory/table/global override plumbing in `aot_runtime.instantiate`.
* Backend-agnostic canon-ABI lift covering `result<_,_>`, `option<T>`, lists, etc.
* AOT loader/compiler retention of imported memory/table/global descriptors.

## Co-authored-by
Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/component/core_backend.zig
+++ b/src/component/core_backend.zig
@@ -127,6 +127,22 @@ pub const PrecompiledCore = struct {
     cwasm_bytes: []const u8,
 };
 
+/// Process-global toggle for AOT debug diagnostics. Off by default;
+/// `main.zig` sets it during startup when the `WAMR_AOT_DEBUG` env var
+/// is set to a non-empty / non-`0` / non-`false` value. Read via
+/// `debugAotEnabled` from any code path that wants to surface AOT
+/// instantiation / call / trap details that would otherwise be
+/// swallowed by the `error.Trap` envelope (see #644).
+var aot_debug_enabled: bool = false;
+
+pub fn setDebugAotEnabled(on: bool) void {
+    aot_debug_enabled = on;
+}
+
+pub fn debugAotEnabled() bool {
+    return aot_debug_enabled;
+}
+
 /// Caller-supplied instantiation options. Today only `precompiled_cores`
 /// is wired; the struct exists so future opts (verbose load logging,
 /// custom AOT host imports, …) can be added without breaking the API.
--- a/src/component/executor.zig
+++ b/src/component/executor.zig
@@ -19,6 +19,8 @@ const HostTrapInfo = @import("../runtime/common/exec_env.zig").HostTrapInfo;
 const interp = @import("../runtime/interpreter/interp.zig");
 const indexspace = @import("indexspace.zig");
 const aot_runtime = @import("../runtime/aot/runtime.zig");
+const core_backend = @import("core_backend.zig");
+const debugAotEnabled = core_backend.debugAotEnabled;
 const Allocator = std.mem.Allocator;
 
 const ComponentInstance = instance_mod.ComponentInstance;
@@ -259,7 +261,13 @@ pub fn callComponentFuncByLocal(
     // backend is AOT-only. (#625 phase 3.)
     if (core_entry.module_inst == null) {
         if (core_entry.aot_inst) |ai| {
-            return callComponentFuncByLocalAot(
+            if (debugAotEnabled()) {
+                std.debug.print(
+                    "[aot-debug] callComponentFuncByLocal -> AOT core_inst_idx={d} core_func_idx={d} func_type_idx={d}\n",
+                    .{ exported.core_instance_idx, exported.core_func_idx, exported.func_type_idx },
+                );
+            }
+            const r = callComponentFuncByLocalAot(
                 owner_inst,
                 ai,
                 exported,
@@ -267,6 +275,14 @@ pub fn callComponentFuncByLocal(
                 out_results,
                 allocator,
             );
+            if (debugAotEnabled()) {
+                if (r) |_| {
+                    std.debug.print("[aot-debug] callComponentFuncByLocal AOT result: ok\n", .{});
+                } else |err| {
+                    std.debug.print("[aot-debug] callComponentFuncByLocal AOT result: {s}\n", .{@errorName(err)});
+                }
+            }
+            return r;
         }
         return error.CoreInstanceNotAvailable;
     }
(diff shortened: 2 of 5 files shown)


cataggar:
Follow-up issues for the remaining AOT-execution work (out of scope for this PR per its description):

* #648 — per-import C-ABI trampoline pool for canon.lower function imports
* #649 — cross-instance memory/table/global import wiring  
* #650 — backend-agnostic canon-ABI lift/lower for compound types

When all three land, the auto-fallback in this PR stops triggering for stdio-echo and similar wit-bindgen P2 components.

Issue #707, opened 2026-05-29
AOT: canon-lower-aot dispatcher rejects compound-type params/results (UnsupportedSignature × 26) → silent zero return → guest exits successfully with no output


cataggar:
## Summary

After #706 (run-on-instantiation) and #704 (resource-builtin bridging) landed, the keyvault repro no longer traps inside the WASIp1 adapter. Instead, **every** canon-lower call the adapter makes into wamr-side WASIp2 host functions is rejected by `dispatchAotComponentTrampoline` with `error.UnsupportedSignature`, the dispatcher returns `{status=1, value=0}`, and `genericDispatcher` swallows that into a `0` return value — so the adapter silently gets `0`/empty everywhere and codegen-cli runs to completion without printing anything (`exit 0`, no output files, no stdout, no stderr after wamr's own banner).

This is the same "trap-stub-as-success" failure mode as #687/#689/#691, but on the canon-lower-AOT path rather than the cross-instance / SignatureTooWide ones.

## Repro (current main, after #704 + #706)

```bash
cd /home/g/wamr && git pull && zig build -Doptimize=ReleaseFast
cd /home/g/azure-sdk-for-zig/codegen/cli/zig-out/bin && \
  rm -rf *.cwasm codegen-cli.composed.cwasm.json && \
  /home/g/wamr/zig-out/bin/wamrc compile-component codegen-cli.composed.wasm
cd /home/g/azure-sdk-for-zig && \
  WAMR=1 codegen/cli/scripts/run.sh \
    /home/g/azure-rest-api-specs/specification/keyvault/data-plane/Secrets /tmp/kv-wamr
# → exit 0, /tmp/kv-wamr is empty, no stdout from codegen-cli
```

(Wasmtime baseline produces a full Zig output tree in /tmp/kv-wasmtime in ~3 s.)

## Diagnostic findings (instrumentation in `wamrAotDispatchComponentTrampolineAot` + each `UnsupportedSignature` arm of `dispatchAotComponentTrampoline`, reverted before filing)

26 canon-lower-aot calls fire and all fail; 9 distinct `(component_func_idx, sig-shape, reject-arm)` tuples:

```
[diag-canon-lower-aot] component_func_idx=43 params=1 results=0 has_retptr=false err=UnsupportedSignature
  → [diag-sig] reg_index mismatch (0 != 1+false)
[diag-canon-lower-aot] component_func_idx=41 params=1 results=0 has_retptr=false err=UnsupportedSignature
[...]

Issue #648, opened 2026-05-21
AOT: per-import C-ABI trampoline pool for canon.lower function imports (#644 follow-up)


cataggar:
Spun out from #644. PR #647 auto-falls-back to interp for any core whose imports the AOT host-bridge can't resolve (notably wit-bindgen-emitted P2 cores that import `wasi:io/streams@0.2.0.…` resolved by component-level `canon.lower` defs). To actually run such cores on AOT we need a per-import dispatch mechanism.

## Approach

Pool of 256 executable stubs, each ~16 bytes:
* x86_64 sysv: `mov r10, <slot_id> ; jmp generic_dispatcher`
* aarch64: `movz x16, <slot_id> ; b generic_dispatcher`

Generic dispatcher signature: `fn(slot, u64, u64, u64, u64, u64, u64) callconv(.c) u64` covers all integer-only signatures up to 6 args via x86_64 sysv ABI uniformity (and aarch64 AAPCS for the first 8 in x0-x7).

The dispatcher looks up the slot's owning component-instance + canon.lower idx, builds an interp call frame, and invokes the underlying lifted func via `componentTrampoline`.

## Files likely touched

* New: `src/runtime/aot/host_trampolines.zig` (asm stub allocator + mprotect).
* `src/runtime/aot/host_bridge.zig`: `resolveAotHostFunction` consults the trampoline pool before falling back to null.
* `src/component/instance.zig`: feasibility probe (`firstUnsupportedAotImport`) accepts canon.lower-resolvable function imports once trampolines are in.

## Acceptance

stdio-echo.wasm's main core runs on AOT (not just via interp fallback). The fallback warning in PR #647 disappears for the function-import case.

## References
* PR #647 (fallback)
* #644 (umbrella)

PR #714, opened 2026-05-29
fix #708: AOT dispatcher loud-by-default warn-once + non-zero sentinel

### LINKED ISSUE #708: AOT dispatcher: silent-zero on failure hides real bugs; loud-by-default warn-once and non-zero sentinel needed
## Summary

Several places in the AOT host-trampoline / dispatcher path swallow real failures and produce a "success-shaped" return value (typically `0`), so a downstream guest sees a wire-OK call that returned an obviously wrong value. Failures that should be loud:

1. **`genericDispatcher` collapsing `status=1` → `value=0`** (`src/runtime/aot/host_trampolines.zig:175-176`). Any dispatcher kind (`canon_lower`, `canon_lower_aot`, `cross_instance`, `canon_builtin_aot`, `trap_stub`) returning `{status=1, value=0}` is delivered to the guest as the u64 `0`. The guest interprets it as e.g. an `own<output-stream>` handle `0`, or an `args_sizes_get` count of `0`, and continues with bogus state.
2. **`canon.lower(aot) trampoline failed` print gated behind `debugAotEnabled()`** (`src/component/executor.zig:6032-6037`). All four dispatcher arms (`Component`, `ComponentAot`, `CrossInstance`, `CanonBuiltin`) silence their `error.UnsupportedSignature` / `error.HostFuncNotBound` paths unless `WAMR_AOT_DEBUG=1` is set. The first time we hit one of these on a real component is silent; debugging requires re-running with a different env var.
3. **`genericDispatcher` silencing slot / pool errors** (`src/runtime/aot/host_trampolines.zig:163-167`). `g_active_pool orelse return 0`, `slot >= pool.next_slot → return 0`, and `entry.ctx orelse return 0` all map "this trampoline slot is broken" to "the host call returned 0". Should at minimum log once per slot before returning the safe-but-wrong value.

The keyvault repro (#707) made this concrete: the WASIp1 adapter calls 26 canon-lower-aot imports, every single one fails with `error.UnsupportedSignature`, every single one returns `0` to the adapter, and codegen-cli runs cleanly to exit 0 with no output and no diagnostic. The silent-zero pattern bit me on **every single intermediate issue** in the sequence #687 / #689 / #691 / #694 / #698 / #705 / #707 — in each case the first-pass debug output looked like "the guest just trapped with unreachable" or "the guest just exited cleanly with no output", and only adding an unconditional print to the dispatcher exposed the actual cause.

### Fix A — change the failure return value

`genericDispatcher` could return a sentinel like `0xDEAD_DEAD_DEAD_DEAD` for `status != 0`. Most guest consumers of these returns will then trip a wit-bindgen `unreachable` (handles, lengths, discriminants are all bounded), surfacing the wamr-side failure as a guest-side trap with a stack pointing into the failing import — *much* easier to debug than silent zero.

There's some history of guest code being tolerant of `0` returns (`args_sizes_get` writes the 0 to its outparams and the caller treats "0 args" as legitimate), so 0 isn't a useful tripwire. A non-zero, non-aligned sentinel will fail on any subsequent pointer use.

### Fix B — log once per slot on failure

Add an `[u8 already_logged]` field (or atomic flag) per `Slot`. On the first `status != 0` for a slot, emit:

```
std.log.warn("[aot] dispatcher slot {d} ({s}) returned status={d}; subsequent calls will be silenced",
    .{ slot, @tagName(entry.dispatch_kind), dispatched.status });
```

Repeat hits for the same slot stay silent so we don't spam stderr (the keyvault repro hits 26 calls on a few hot slots).

### Fix C — drop `debugAotEnabled()` gates on dispatcher failure prints

The four dispatcher arms have prints like:

```zig
if (debugAotEnabled()) {
    std.debug.print("[aot-dispatch] canon.lower(aot) trampoline failed: {s}\n", .{@errorName(err)});
}
```

Replace with `std.log.warn` (which is rate-limited by the logging facility, not by the user's env). The failure shape is small (single line) and we'd only print on the *failing* path, not every call.

### Fix D — surface dispatcher-slot lookup failures

The three early returns in `genericDispatcher` (`g_active_pool == null`, `slot >= next_slot`, `entry.ctx == null`) all indicate a wamr-internal state-machine bug — they should never fire in normal operation. Log them at `error` level and ideally `@panic`, so they show up immediately:

```zig
const pool = g_active_pool orelse {
    std.log.err("[aot-dispatcher] called with no active TrampolinePool (slot={d})", .{slot});
    return 0;
};
if (slot >= pool.next_slot) {
    std.log.err("[aot-dispatcher] slot {d} >= next_slot {d}", .{ slot, pool.next_slot });
    return 0;
}
```

### Fix E (secondary, related) — `WAMR_AOT_DEBUG=1` segfaults on this repro

With the canon-lower-aot dispatcher silently failing 26 times, running the keyvault bundle works (exit 0, no output). Re-running with `WAMR_AOT_DEBUG=1` — which only adds a `std.debug.print` per failing dispatch — **segfaults** the host:

```
$ WAMR=1 codegen/cli/scripts/run.sh <spec> <out>
…
exit: 0    # silent failure
$ WAMR_AOT_DEBUG=1 WAMR=1 codegen/cli/scripts/run.sh <spec> <out>
…
[aot-dispatch] canon.lower(aot) trampoline failed: UnsupportedSignature
…
Segmentation fault (core dumped)
exit: 139
```

Likely a stack-corruption / write-after-free in the failing trampoline path that's load-bearing-by-luck when no print is emitted. Could be that the `errdefer` cleanup for one of the per-call allocations doesn't run when the `catch` block writes to stderr, or it's an interaction between trampoline-pool RWX pages and a longjmp-style unwind that doesn't take the stderr-buffered lock cleanly. Worth a `valgrind`/`asan` pass over the failing dispatcher path even if the underlying signature fix from #707 makes it moot in practice.

## Acceptance

- Any new "AOT path doesn't yet handle signature X" bug surfaces with a single human-readable wamr-side line on its first occurrence, not as "the guest crashed mysteriously".
- A new test fixture asserts that an `UnsupportedSignature` from `wamrAotDispatchComponentTrampolineAot` triggers the warn-once log and that the dispatched return value is the chosen sentinel (or a recognisable non-zero pattern).
- `WAMR_AOT_DEBUG=1` no longer crashes when the failure rate is high.

## Source pointers

- `src/runtime/aot/host_trampolines.zig:162-177` — `genericDispatcher` (the swallow site)
- `src/component/executor.zig:5993-6037` — `wamrAotDispatchComponentTrampoline` / `wamrAotDispatchComponentTrampolineAot` (the gated prints)
- `src/component/executor.zig:6066-6090` — `wamrAotDispatchCrossInstance` (same gating)
- `src/component/executor.zig:6177-6186` — `wamrAotDispatchCanonBuiltin` (same gating)
- `src/component/executor.zig:6259-6275` — `wamrAotDispatchTrapStub` (already gated; first occurrence here too)
- `src/component/core_backend.zig:147-160` — `debugAotEnabled()` / `setAotDebugEnabled()` plumbing

## Notes

- This is purely a developer-ergonomics / observability issue. Once #707 lands, the 26 failures disappear and the silent-zero won't bite *this* component anymore. But future canon-lower / cross-instance / canon-builtin gaps will hit the same wall — the rule should be "wamr never silently returns 0 for a failed dispatch on first occurrence".
- The keyvault repro is reproducible against `cataggar/azure-sdk-for-zig` main and provides a high-failure-rate test for the proposed warn-once logic.

cataggar:
Closes #708.

Before, every AOT dispatcher arm (`canon_lower`, `canon_lower_aot`, `cross_instance`, `canon_builtin_aot`, `trap_stub`) collapsed its `UnsupportedSignature` / `HostFuncNotBound` / `OutOfMemory` failure mode into a guest-visible `u64 = 0`, and the per-arm diagnostic prints were gated behind `WAMR_AOT_DEBUG=1`. The first time a real component hit a missing host wiring (most recently #707's keyvault repro with 26 silenced `UnsupportedSignature` failures) the run looked like *guest exited cleanly with no output*, and only adding an unconditional print to the dispatcher exposed the actual cause.

## What this PR does

- **Fix A** — Adds `DISPATCH_FAILURE_SENTINEL = 0xDEAD_DEAD_DEAD_DEAD` in `host_trampolines.zig` and returns it from `genericDispatcher` whenever a dispatcher arm reports `status != 0` or the slot-lookup state machine has an invariant violation. The sentinel is non-zero, non-aligned, and outside every plausible canon-ABI small-integer range (handles, lengths, discriminants), so the first downstream use trips a wit-bindgen `unreachable` and produces a stack pointing at the failing import — versus the silent zero that used to thread several more operations before crashing somewhere unrelated.
- **Fix B** — Adds `already_logged_failure: bool` to `Slot` and emits a single `std.log.warn` line the first time a slot's dispatcher arm reports `status != 0`. Subsequent failures from the same slot stay silent so high-failure-rate repros (the keyvault bundle hits the same hot slots 26 times) don't spam stderr.
- **Fix C** — Drops the four `if (debugAotEnabled())` wrappers around the dispatcher-arm failure prints in `executor.zig` (canon_lower, canon_lower_aot, cross_instance, canon_builtin_aot) plus the trap_stub one, and switches `std.debug.print` to `std.log.warn`. Diagnostics now go through the logging facility and respect the configured `log_level`, instead of being silenced by default until the user re-runs with an env var. `debugAotEnabled()` itself stays — other instantiation-trace call sites still read it.
- **Fix D** — Replaces the three `orelse return 0` / bounds-check `return 0` arms in `genericDispatcher` with `std.log.warn` lines explaining which invariant fired (no active pool, OOR slot, null ctx), and returns the sentinel. The issue body suggested `std.log.err`; pragmatically I used `warn` so existing tests that deliberately exercise the null-ctx fallback (e.g. `#648 phase 1`) don't trip Zig's test runner's err-log-counts-as-failure rule.
- **Tests** — New cases in `aot_host_trampolines_test.zig` assert that `genericDispatcher` returns `DISPATCH_FAILURE_SENTINEL` on a failing host-fn call, that the per-slot warn-once flag flips after the first failure (so the second call is silent), and that the no-active-pool / OOR-slot paths also return the sentinel. The pre-existing test at `aot_host_trampolines_test.zig:115` was updated to expect the sentinel from a null-ctx slot (the same path that assertion was already exercising; previously expected `0`).

## What's NOT in this PR

**Fix E** from the issue (segfault under `WAMR_AOT_DEBUG=1` on the keyvault repro) is out of scope here — the repro lives in `cataggar/azure-sdk-for-zig`, not this repo, and once #707's signature gap closes the failure path executes far less often. **A follow-up issue should be opened once a smaller in-repo repro exists.** The suspected `errdefer`-not-running / longjmp-vs-stderr-lock theory from #708's issue body is worth a valgrind/asan pass over the failing dispatcher path.

## Validation

```
unset ZIG_LOCAL_CACHE_DIR
export ZIG_GLOBAL_CACHE_DIR="$PWD/.zig-global-cache"
zig build test --summary all
# Build Summary: 67/67 steps succeeded; 2945/2952 tests passed (7 skipped)
```

The new tests intentionally emit `std.log.warn` lines on the failure paths they exercise — those appear in stderr during the test run but do not fail the suite (Zig's test runner only counts `.err`-level logs).

### DIFF
--- a/src/runtime/aot/host_trampolines.zig
+++ b/src/runtime/aot/host_trampolines.zig
@@ -46,6 +46,21 @@ pub const DispatchResult = extern struct {
     err_name: ?[*:0]const u8 = null,
 };
 
+/// Sentinel u64 returned to the guest by `genericDispatcher` when a
+/// dispatcher arm reports `status != 0` (or when the slot-lookup
+/// state machine has hit an invariant violation). Chosen to be
+/// non-zero, non-aligned, and outside every plausible canon-ABI
+/// "small integer" range (handles, lengths, discriminants) so that
+/// the *first* downstream use in guest code trips a wit-bindgen
+/// `unreachable`, producing a stack trace pointing at the failing
+/// import. Returning `0` instead — which we did until #708 — looks
+/// like a legitimate empty handle / zero-length buffer / first
+/// discriminant arm, so the guest happily threads the failure
+/// through several more operations before crashing somewhere
+/// unrelated. Pair-with the per-slot warn-once line emitted by
+/// `genericDispatcher` for the operator-side context.
+pub const DISPATCH_FAILURE_SENTINEL: u64 = 0xDEAD_DEAD_DEAD_DEAD;
+
 extern fn wamrAotDispatchComponentTrampoline(
     ctx_opaque: *anyopaque,
     lowered_sig: *const LoweredSig,
@@ -178,11 +193,29 @@ pub fn setActivePool(pool: ?*TrampolinePool) void {
 var g_dispatch_warn_once: std.bit_set.IntegerBitSet(MAX_SLOTS) = .initEmpty();
 
 pub fn genericDispatcher(slot: u32, a0: u64, a1: u64, a2: u64, a3: u64, a4: u64, a5: u64, a6: u64, a7: u64, a8: u64, a9: u64) callconv(.c) u64 {
-    const pool = g_active_pool orelse return 0;
-    if (slot >= pool.next_slot) return 0;
+    const pool = g_active_pool orelse {
+        std.log.warn(
+            "[aot dispatch] called with no active TrampolinePool (slot={d}); returning sentinel 0x{x:0>16}",
+            .{ slot, DISPATCH_FAILURE_SENTINEL },
+        );
+        return DISPATCH_FAILURE_SENTINEL;
+    };
+    if (slot >= pool.next_slot) {
+        std.log.warn(
+            "[aot dispatch] slot={d} out of range (next_slot={d}); returning sentinel 0x{x:0>16}",
+            .{ slot, pool.next_slot, DISPATCH_FAILURE_SENTINEL },
+        );
+        return DISPATCH_FAILURE_SENTINEL;
+    }
 
     const entry = pool.slots[slot];
-    const ctx = entry.ctx orelse return 0;
+    const ctx = entry.ctx orelse {
+        std.log.warn(
+            "[aot dispatch] slot={d} has null ctx (kind={s}); returning sentinel 0x{x:0>16}",
+            .{ slot, @tagName(entry.dispatch_kind), DISPATCH_FAILURE_SENTINEL },
+        );
+        return DISPATCH_FAILURE_SENTINEL;
+    };
     const dispatched = switch (entry.dispatch_kind) {
         .canon_lower => wamrAotDispatchComponentTrampoline(ctx, &entry.lowered_sig, a0, a1, a2, a3, a4, a5, a6, a7, a8, a9),
         .canon_lower_aot => wamrAotDispatchComponentTrampolineAot(ctx, &entry.lowered_sig, a0, a1, a2, a3, a4, a5, a6, a7, a8, a9),
@@ -191,10 +224,10 @@ pub fn genericDispatcher(slot: u32, a0: u64, a1: u64, a2: u64, a3: u64, a4: u64,
         .trap_stub => wamrAotDispatchTrapStub(ctx, &entry.lowered_sig, a0, a1, a2, a3, a4, a5, a6, a7, a8, a9),
     };
     if (dispatched.status != 0) {
-        // Surface the silent-zero collapse once per slot. The `trap_stub`
-        // kind is an explicit, expected trap path — it has already
-        // printed its own diagnostic and the slot pre-emptively means
-        // "this import is known-unbound", so no extra warning helps.
+        // Surface the failure once per slot. The `trap_stub` kind is an
+        // explicit, expected trap path — it has already printed its own
+        // diagnostic and the slot pre-emptively means "this import is
+        // known-unbound", so no extra warning helps.
         if (entry.dispatch_kind != .trap_stub and slot < MAX_SLOTS and !g_dispatch_warn_once.isSet(slot)) {
             g_dispatch_warn_once.set(slot);
             const err_str: []const u8 = if (dispatched.err_name) |p|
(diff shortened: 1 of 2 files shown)


cataggar:
Rebased onto current `main` (`b9f2fc76`). Scope of this PR shrank significantly because #711 landed the per-slot warn-once mechanism (`g_dispatch_warn_once` bitset) and added `DispatchResult.err_name` in parallel:

- **Fix A (sentinel)** — still here; this is the main remaining contribution.
- **Fix B (warn-once)** — already landed via #711's `g_dispatch_warn_once`. Dropped my redundant `Slot.already_logged_failure` field.
- **Fix C (drop `debugAotEnabled()` gates)** — implicitly addressed by #711: the per-arm catches now populate `err_name` and the central warn-once line surfaces it in release builds. Dropped my edits to the per-arm prints.
- **Fix D (loud state-machine warns)** — still here.
- **Fix E** — deferred (repro outside this repo).

Net diff is now `+120 / -11` across 2 files (down from the original `+165 / -34` across 3 files).

Issue #743, opened 2026-06-01
AOT: deeper tcgc.compile trap surfaces past #719 — OOB at local_func[11040]+0x33c3 (mem_size=0x1df0000)


cataggar:
## Summary

With the #719 fix chain (#721, #723, #725, #727, #730, #731, #732) merged, the keyvault repro from #719 now gets MUCH further into `tcgc.compile` before tripping a different, deeper trap:

```
wamr: loaded AOT manifest from …/codegen-cli.composed.cwasm.json (8 cores precompiled)
collecting spec files from /spec
…(stdlib collection lines)…
emitter-options size = 1345719 bytes
calling tcgc.compile...
wasm trap: out of bounds memory access (code+0x2800beb, local_func[11040]+0x33c3, mem_size=0x1df0000)
```

Wasmtime baseline runs the same bundle end-to-end successfully (`got 58187 bytes back from tcgc.compile`).

## Differences from #719

| | #719 (FIXED) | this issue |
|---|---|---|
| trap kind | `unreachable` | `out of bounds memory access` |
| `local_func` | 26 | 11040 |
| code offset | 0x950 | 0x2800beb |
| mem_size at trap | not reported | 0x1df0000 (~30.6 MB) |
| timing | very early in tcgc.compile | very deep |

## Repro

```bash
cd /home/g/wamr && git pull && zig build -Doptimize=ReleaseFast   # head must include #732
cd /home/g/azure-sdk-for-zig/codegen/cli/zig-out/bin && \
  rm -rf *.cwasm codegen-cli.composed.cwasm.json && \
  /home/g/wamr/zig-out/bin/wamrc compile-component codegen-cli.composed.wasm
cd /home/g/azure-sdk-for-zig && \
  WAMR=1 codegen/cli/scripts/run.sh \
    /home/g/azure-rest-api-specs/specification/keyvault/data-plane/Secrets /tmp/kv-wamr
# → trap: OOB @ local_func[11040]+0x33c3, mem_size=0x1df0000
```

## Hypotheses (ordered by likelihood)

[...]

PR #776, opened 2026-06-03
compiler/x86_64: model helper call clobbers

cataggar:
## Summary
- Fix the large memory.fill REP STOSB clobber mask to use the actual rdi allocator index.
- Mark runtime-helper IR ops as call-clobber points for x86_64 register allocation.
- Add a regression for the rdi clobber mask.

Refs #743.

## Validation
- zig build test --summary all
- Keyvault default repro no longer hits the original local_func[11040]+0x33c3 OOB; it progresses to a later local_func[10677] guest unreachable, tracked on #743.


cataggar:
Local validation update: this fixes the original keyvault OOB (`local_func[11040]+0x33c3`) on the #743 repro, but the issue should remain open after merge. With this branch the same default repro progresses to a later explicit guest allocator abort (`local_func[10677]+0x6f2`), while `compile-component -O0` still succeeds. I updated #743 with the follow-up probes.

PR #779, opened 2026-06-05
compiler: make core4 optimized AOT compile finish (codegen liveness + optimizer scoping) (#778)

### LINKED ISSUE #778: Optimized AOT compile: native codegen is the bottleneck after optimizer O(n^2) fixes (core4)
## Summary

After fixing the optimizer's O(n²) hotspots in #777 (tailDuplicateSmallJoins,
the eqz/sign-extend/select folds, and promoteLocalsToSSA), the **IR optimization
phase now completes** for the 12,610-function `codegen-cli` core4 component core.
The next bottleneck preventing the optimized AOT compile from finishing is
**native codegen** (instruction selection / register allocation / emit), plus two
residual amplifiers in the optimizer. This issue tracks closing the remaining gap
so `wamrc compile` (optimized, verifier off) produces a `.cwasm` in a reasonable
time.

Repro module: core4 of `codegen-cli.composed.wasm` (33.7 MB, 12,610 functions),
unbundled to a standalone `.wasm`. Command:
`wamrc compile --no-verify-ir <core4>.wasm -o out.cwasm` with `WAMR_AOT_VERIFY_IR=off`.

## Current measurements (post-#777, ReleaseFast, x86_64 host, 8 cores)

Full no-timing completion run (3600 s cap):
- Optimization phase **completes**: both module outer passes
  (`inlineSmallFunctions outer 0 = 2764`, `outer 1 = 647`,
  `Optimization: 153454 passes made changes`) in roughly the first ~2,300 s.
- **Native codegen then ran ~1,300 s without finishing → hit the 3600 s cap; no
  `.cwasm` was emitted.**

So codegen is now the dominant remaining cost. (Before #777, optimization itself
never finished — it timed out at `local_func=10,652/12,610`.)

## Bottleneck 1 (primary): native codegen on monster functions

The worst function (`local_func=11396`) has 10,900 blocks and balloons to **874k
IR instructions** after `lowerPhisToLocals` (an inherent out-of-SSA expansion:
each phi with K predecessor edges becomes K copies). Native instruction selection
+ register allocation on functions of this size is the suspected super-linear
cost. There is currently **no codegen-phase timing instrumentation** to attribute
it.

Suggested work:
- Add codegen timing knobs mirroring `WAMR_AOT_PASS_TIMING*` (per-function,
  per-sub-phase: isel / regalloc / scheduling / emit) to localize the cost.
- Profile register allocation specifically on `local_func=11396`; check for
  O(n²)/O(n·live) behavior (interference graph, liveness, spill loops).
- Consider whether critical-edge splitting (adds blocks) vs. the current copy
  expansion reduces post-lowering instruction count, and/or a size-based
  strategy switch for monster functions.

## Bottleneck 2: optimizer module-level 2× outer loop

`runPassesWithOptions` uses `outer_max = 2` (passes.zig) — the whole per-function
pipeline runs over every function twice to let cross-function inlining
(`inlineSmallFunctions`) settle. The 2nd pass largely re-confirms fixpoint at
~full cost. Worth investigating whether the 2nd module pass can be scoped to only
functions affected by outer-0 inlining (647 inlines in outer 1) instead of all
12,610.

## Bottleneck 3: residual per-function optimization cost on large `>cap` functions

Functions above the tail-dup block cap still cost seconds each, dominated by
repeated `globalValueNumbering` + `computeDominators` + `forwardRedundantLoadsDominator`
across inner fixpoint iterations (each call is sub-second, i.e. not O(n²), but
they run many times). Top remaining optimization-phase function totals (outer 0):

| local_func | blocks | insts | total opt time |
|-----------:|-------:|------:|---------------:|
| 11396 | 10,900 | 82,267 | 58.2 s |
| 11780 |  8,598 | 62,642 | 21.6 s |
| 11740 |  8,336 | 67,662 | 17.6 s |
| 10760 |  4,794 | 35,164 | 14.4 s |
|  6145 |  3,746 | 30,982 | 12.1 s |
| 11394 |  3,310 | 31,073 | 12.0 s |

Possible: cache/skip dominator recomputation when the CFG is unchanged between
inner iterations (a CFG analysis cache already exists — verify it is being reused
here), and/or memoize GVN value numbers across iterations.

## Acceptance

- Optimized `wamrc compile --no-verify-ir` of core4 produces a `.cwasm` well within
  a practical CI budget.
- Ideally the full `compile-component` (all cores) + the keyvault end-to-end
  generation finishes without timeout.

cataggar:
Closes #778.

Makes the optimized AOT compile of the 12,610-function `codegen-cli` core4
component core actually finish and emit a `.cwasm` in a practical time.
Before: optimization ran ~2,300 s then native codegen hit the 3,600 s cap with
**no output**. After: the full compile finishes in **~60 s** and emits a valid
86.5 MB `.cwasm`.

### What changed
1. **Codegen timing instrumentation** (`WAMR_AOT_CODEGEN_TIMING*`) — mirrors
   `WAMR_AOT_PASS_TIMING*` for the native-codegen phase: per-function
   setup/liveness/regalloc/emit partition + `hashFunction` time + a module
   summary. Disabled by default (hard no-op). This **disproved the issue's
   regalloc hypothesis**: register allocation is <50 ms even on the monster
   functions — the hotspot is **liveness**.
2. **Worklist liveness solver** — `computeLiveness` was a round-robin
   iterate-until-stable sweep over blocks in reverse index order, rescanning
   every block each round. Replaced with a predecessor-driven worklist that
   only revisits a block when a successor's `live_in` grew. Converges to the
   identical least fixpoint (same transfer function, same monotone lattice), so
   it is behavior-preserving by construction. The old solver is kept verbatim as
   `computeLivenessRoundRobin` and pinned by a differential test.
3. **Scope the optimizer's 2nd outer pass** — `runPassesWithOptions` re-ran the
   full per-function pipeline over all 12,610 functions on the 2nd outer round.
   Now it only revisits functions that inlining actually changed in that round;
   untouched functions are provably already at fixpoint.
4. **CFG analysis cache reuse** — `invalidate()` is now lazy: the next lookup
   compares cached successor lists to the current function and reuses
   successors/predecessors/dominators when the CFG is unchanged, instead of
   recomputing every inner fixpoint iteration.

### Results (ReleaseFast, x86_64)
- core4 optimized compile: **3,600 s cap / no output → ~60 s** + valid `.cwasm`.
- Codegen total 44.6 s → 33.5 s (liveness worklist); worst-function liveness up
  to **4.2×** faster (local_func=10760: 6.16 s → 1.46 s).
- **Strongest validation:** recompiling core4 with the worklist liveness yields a
  **byte-identical** `.cwasm` (sha `2ec5678c…`) vs the round-robin baseline.

### Tests
- Full suite green (`zig build test --summary all`): 79/79 steps,
  3,653/3,660 passed, 7 skipped, 0 failures.
- New: `codegenTimingOptionsFromEnv` parsing, a non-invasive byte-identical
  codegen-timing test, `shouldLogFunc`/`FuncTimer` gating, a scoped-vs-unscoped
  deep-IR-equality test, CFG-cache reuse/recompute tests, and a worklist-vs-
  round-robin differential liveness test across diamond/loop/nested-loop/chain
  CFGs.

### Not in scope (follow-ups)
Emit is now the largest single codegen cost on high-block functions
(local_func=11396 emit ≈6.6 s) — filed separately. aarch64 reports per-function
totals + module summary only (no sub-phase breakdown).

### DIFF
--- a/src/compiler/codegen/aarch64/compile.zig
+++ b/src/compiler/codegen/aarch64/compile.zig
@@ -11,6 +11,8 @@ const schedule = @import("schedule.zig");
 const coalesce_post = @import("coalesce_post.zig");
 const range_split = @import("../../ir/range_split.zig");
 const codegen_cache = @import("../../codegen_cache.zig");
+const passes = @import("../../ir/passes.zig");
+const codegen_timing = @import("../timing.zig");
 
 /// Compile-time debug flag: when true, print per-function range-split
 /// stats to stderr. Flip via `zig build -Drange-split-debug=true` after
@@ -556,6 +558,13 @@ pub const CompileOptions = struct {
     /// follow-up. Disabling this option also disables the scan on the
     /// `coalesce_post.zig` unit tests via build.zig's test step.
     enable_post_emit_coalesce: bool = true,
+    /// #778: native-codegen timing diagnostics. aarch64 reports per-function
+    /// totals + a module summary (no x86_64-style setup/liveness/regalloc
+    /// sub-phase breakdown — the #778 bottleneck is the x86_64 host path).
+    codegen_timing: passes.CodegenTimingOptions = .{},
+    /// Component core/module index, used only to match the
+    /// `WAMR_AOT_CODEGEN_TIMING_MODULE` filter.
+    module_idx: u32 = 0,
 };
 
 /// Context threaded through per-function compilation for cross-function
@@ -8371,11 +8380,31 @@ pub fn compileModuleCachedWithOptions(
 
     var stats: codegen_cache.CacheStats = .{};
 
+    // #778: codegen timing (coarse — per-function totals + module summary).
+    const ct = options.codegen_timing;
+    const ct_live = ct.enabled and ct.moduleMatches(options.module_idx);
+    const module_start_ns: u64 = if (ct_live) codegen_timing.nowNs() else 0;
+    var mod_hash_ns: u64 = 0;
+    var mod_compile_ns: u64 = 0;
+    var mod_compiled: usize = 0;
+    var mod_reused: usize = 0;
+    codegen_timing.printModuleBegin(ct, options.module_idx, func_count);
+
     for (ir_module.functions.items, 0..) |func, fi| {
         const func_base: u32 = @intCast(all_code.items.len);
         try offsets.append(allocator, func_base);
 
-        const ir_sha = codegen_cache.hashFunction(&func);
+        var hash_ns: u64 = 0;
+        const ir_sha = blk: {
+            if (ct_live) {
+                const t0 = codegen_timing.nowNs();
+                const sha = codegen_cache.hashFunction(&func);
+                hash_ns = codegen_timing.nowNs() -| t0;
+                mod_hash_ns += hash_ns;
+                break :blk sha;
+            }
+            break :blk codegen_cache.hashFunction(&func);
+        };
 
         var reused = false;
         var hit_code: []const u8 = undefined;
@@ -8392,6 +8421,7 @@ pub fn compileModuleCachedWithOptions(
 
         var cache_code_owned: []u8 = undefined;
         var cache_patches_owned: []codegen_cache.FuncCallPatch = undefined;
+        var compile_ns: u64 = 0;
 
         if (reused) {
             cache_code_owned = try allocator.dupe(u8, hit_code);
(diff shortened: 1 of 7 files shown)


cataggar:
Follow-up issues filed (out of scope for this PR):
- #780 — emit phase is super-linear in block count on large functions (now the #1 codegen cost)
- #781 — add aarch64 codegen timing sub-phase breakdown
- #782 — further liveness speedup (worklist seed ordering / bitset live-sets)

PR #788, opened 2026-06-05
compiler/ir: faster liveness via RPO seed order + skip + scratch reuse (#782)

### LINKED ISSUE #782: AOT codegen: further liveness speedup (worklist seed ordering / bitset live-sets)
Follow-up from #778 / #779.

#779 replaced the round-robin liveness solver with a predecessor-driven worklist
(`computeLiveness`, `src/compiler/ir/analysis.zig`), cutting liveness up to 4.2×
and keeping the core4 `.cwasm` byte-identical. The worklist is currently **seeded
in plain block-index order**; liveness on the worst function is still ~4.2 s
(local_func=11396, 10,900 blocks).

### Suggested work (further, lower-priority speedup)
- Seed the worklist in **reverse-postorder of the reverse CFG** (i.e. process a
  block after its successors) so the first sweep propagates further before
  re-enqueues — fewer total block visits to converge.
- Consider a **dense bitset** live-set representation (vregs are dense `0..next_vreg`)
  instead of `AutoHashMap(VReg, void)` where live sets are large, trading memory
  for faster union/diff. Watch peak memory on 10k-block functions.
- Must remain behavior-preserving: keep `computeLivenessRoundRobin` as the
  differential oracle, and verify core4 `.cwasm` stays byte-identical
  (sha `2ec5678c…`).

cataggar:
Closes #782. Follow-up to #778/#779.

Three behavior-preserving speedups to the `computeLiveness` worklist solver
(`src/compiler/ir/analysis.zig`) — the AOT register allocator's backward
liveness dataflow that dominates codegen on functions with thousands of blocks
(worst function ~4.2 s: 10,900 blocks, ~874k instructions after
`lowerPhisToLocals`).

### What changed
1. **Reverse-postorder seed ordering.** The worklist was seeded in plain
   block-index order. It is now seeded so the LIFO worklist pops blocks in
   forward-CFG **postorder** (each block after its successors) via a new
   iterative `forwardPostorder` helper — the convergence-optimal visit order
   for a backward dataflow, minimizing total block visits. A multi-start DFS
   covers unreachable components deterministically.
2. **Skip the transfer walk when `live_out` did not grow.** `live_in` depends
   only on `live_out` and the block's fixed use/def, so once a block has been
   visited, an unchanged `live_out` means `live_in` cannot change — the
   backward instruction walk is skipped. This avoids redundant transfer
   recomputation on re-enqueues (the largest single lever on the monster
   functions).
3. **Reuse one scratch live set** across visits (`clearRetainingCapacity`)
   instead of allocating/freeing an `AutoHashMap` per block visit, removing
   allocator churn on the hot path.

Seed/visit order only affects convergence speed, **never the least fixpoint**,
so the change is behavior-preserving by construction.

### Dense bitset: intentionally not adopted
The issue's optional dense-bitset live-set idea is **not** taken. Live sets are
sparse (most vregs die in their defining block); a per-block dense bitset over
`0..next_vreg` would reach GB-scale on the very 10k-block / ~874k-instruction
function it targets. The existing sparse `AutoHashMap` representation is kept.

### Validation
- `computeLivenessRoundRobin` is kept verbatim as the differential oracle.
- The differential test (`worklist == round-robin`) is extended with
  **unreachable-block, unreachable-into-reachable, multi-exit, and self-loop**
  CFG shapes to exercise the new seed path.
- Full suite green (`zig build test`): **79/79 steps, 3658/3666 passed, 8
  skipped, 0 failures**.
- Set fixpoint is identical and consumers treat live sets as order-independent,
  so core4 `.cwasm` is expected to remain byte-identical (sha `2ec5678c…`);
  not re-run locally (no core4 wasm in this environment).

### DIFF
--- a/src/compiler/ir/analysis.zig
+++ b/src/compiler/ir/analysis.zig
@@ -351,38 +351,72 @@ pub fn computeLiveness(
     defer allocator.free(in_queue);
     @memset(in_queue, false);
 
+    // Tracks whether a block's transfer has run at least once. The first
+    // visit must always recompute live_in (to seed use[B]); on later visits
+    // an unchanged live_out means live_in cannot change, so the transfer
+    // walk can be skipped (issue #782).
+    var visited = try allocator.alloc(bool, nblocks);
+    defer allocator.free(visited);
+    @memset(visited, false);
+
     var worklist: std.ArrayList(ir.BlockId) = .empty;
     defer worklist.deinit(allocator);
-    // Seed every block, low index first; LIFO pop then visits high index
-    // first, matching the old reverse-index sweep's first pass.
-    for (0..nblocks) |seed| {
-        const bid: ir.BlockId = @intCast(seed);
-        try worklist.append(allocator, bid);
-        in_queue[bid] = true;
+    // Seed in reverse-postorder of the forward CFG so the LIFO worklist pops
+    // blocks in postorder — each block after its successors — the order that
+    // lets a backward-liveness sweep propagate furthest before re-enqueues,
+    // minimizing total block visits to converge (issue #782). Seed order
+    // only affects convergence speed, never the fixpoint (pinned by the
+    // `computeLivenessRoundRobin` differential test).
+    {
+        const postorder = try forwardPostorder(func, allocator, &successors);
+        defer allocator.free(postorder);
+        var i: usize = postorder.len;
+        while (i > 0) {
+            i -= 1;
+            const bid = postorder[i];
+            try worklist.append(allocator, bid);
+            in_queue[bid] = true;
+        }
     }
+
+    // Single reusable scratch set for the per-block transfer, cleared (not
+    // freed) each visit to avoid allocator churn on the hot path (issue #782).
+    var live = std.AutoHashMap(ir.VReg, void).init(allocator);
+    defer live.deinit();
 
     while (worklist.pop()) |bid| {
         in_queue[bid] = false;
         const block = &func.blocks.items[bid];
         const bl = liveness.getPtr(bid).?;
 
-        // live_out ∪= live_in[succ]  (monotone accumulate)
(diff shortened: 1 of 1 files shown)


cataggar:
## Measured speedup

Benchmarked this PR's `computeLiveness` against the pre-PR worklist (`origin/main`), ReleaseFast on this aarch64 host, best of 5 reps on large synthetic CFGs (K=128 values live across):

| shape | blocks | old (ms) | new (ms) | speedup |
|---|--:|--:|--:|--:|
| forward chain (favorable numbering) | 10,000 | 75.2 | 63.4 | **1.19×** |
| reverse chain (adversarial numbering) | 10,000 | 3422.3 | 62.9 | **54×** |
| diamond ladder | 9,001 | 70.7 | 59.9 | **1.18×** |

**Takeaways**
- When block numbering is already (reverse-)topological — the lucky case the old index-order seeding implicitly assumed — the gain is a steady **~1.2×**, purely from the *skip-transfer-when-`live_out`-unchanged* and *scratch-map reuse* changes.
- When numbering is unfavorable — which can happen after the optimizer reorders / splits / merges blocks — the old solver degrades to near-O(n²) re-enqueues (~3.4 s here), while RPO seeding stays at ~63 ms: a **54×** win.
- The new solver is essentially **numbering-independent** (62.9 ms vs 63.4 ms across both chains), so it removes the pathological worst case rather than just shaving the average.

For the issue's real ~4.2 s worst function (10,900 blocks, ~874k instructions) the realized gain depends on how topological its post-optimization numbering is, but the new solver guarantees it can no longer hit the pessimal path. That exact function wasn't re-measured locally (no core4 wasm in this environment); correctness is pinned by the extended `computeLivenessRoundRobin` differential test.

<sub>Benchmark harness was a throwaway exe comparing both solvers on the same synthetic `IrFunction`s; not committed.</sub>

Issue #813, opened 2026-06-07
wamrc: IR-pass pipeline pathologically slow on the huge core-4 function (build-layout-sensitive); surfaced via #393 iconst change


cataggar:
## Summary
While implementing #393 zero-extension elimination (PR #812), switching the `iconst_32` x86_64 codegen from the 7-byte REX.W form (`mov r64, imm32`, `48 C7 /0 id`) to the equivalent 5-byte `mov r32, imm32` (`B8+rd id`, which zero-extends) **deterministically hangs `wamrc` codegen on the keyvault/TCGC component** — ~315 s of *measured CPU* (utime), no completion — even though the change is correct (full test suite + CoreMark pass). Reverting just the `iconst` encoding (keeping the rest of #812) compiles the same component in **49 s**.

This is almost certainly a **pre-existing latent bug** that is *triggered by*, not caused by, the shorter/different `iconst` encoding (instruction size/encoding sensitivity), so it's worth fixing independently before the `iconst` zero-ext lever can land.

## Repro
On top of PR #812 (`perf/x86-zext-elim`), apply:

```zig
// emit.zig — add:
pub fn movRegImm32Zx(self: *CodeBuffer, dst: Reg, imm: u32) !void {
    if (dst.isExtended()) try self.emitByte(0x41); // REX.B
    try self.emitByte(0xB8 + @as(u8, dst.low3()));
    try self.emitU32(imm);
}

// compile.zig — in `.iconst_32 => |val|`, replace
//   try code.movRegImm32(dr, val);  + writeDefTyped(...,.i32)
// with
   try code.movRegImm32Zx(dr, @bitCast(val));   // B8+rd, zero-extends
   try writeDef(code, alloc_result, dest, dr);  // no separate zero-extend
```

Then:
```sh
zig build -Doptimize=ReleaseFast
# keyvault composed component (azure-sdk-for-zig codegen-cli.composed.wasm):
wamrc compile-component -o /tmp/kv.cwasm.json <codegen-cli.composed.wasm>
#   -> hangs in core-0 codegen (~315 s CPU, never completes)
```

- The encoding itself is correct: `B8+rd id` is a valid `mov r32, imm32` and the CRCs/output are byte-identical where it *does* finish (CoreMark, the full `zig build test` suite).
- It is **not contention**: `utime ≈ elapsed ≈ 315 s` (a busy spin), and the same build with `iconst` reverted finishes in 49 s back-to-back.
[...]

PR #816, opened 2026-06-07
fix(ir): converge the per-function pass fixpoint on pathological functions (#813)

cataggar:
Fixes the build-layout-sensitive, pathologically slow IR pass pipeline reported in #813.

The per-function optimization fixpoint (`runPassesWithOptionsScoped`, cap 8) could run to its iteration cap **without converging**, producing build-layout-sensitive codegen and, on the keyvault component's monster function, multi-minute compiles. This PR fixes two independent non-convergence sources.

## 1. `hoistLoopInvariantCode` ⇄ `tailDuplicateSmallJoins` preheader fight (root cause)
`hoistLoopInvariantCode` synthesized a loop preheader **eagerly** for every loop — even loops with nothing to hoist — creating an empty `br header` block. `tailDuplicateSmallJoins` (which has no minimum-body guard) then **absorbs** that empty preheader, restoring the multi-entry loop. The next outer iteration re-synthesizes the preheader, and the two passes **ping-pong forever** (leaking one dead block per iteration). So the fixpoint always hits the cap and the emitted code depends on *where the cap truncates* — the "build-layout-sensitive" behavior in #813.

**Fix:** synthesize the preheader **lazily**, only on the first instruction actually hoisted.

Reproducible on **CoreMark** (`local_func=26`):
| metric | before | after |
|---|---|---|
| outer fixpoint iterations | 8 (capped, never converges) | **3 (converges)** |
| output vs outer cap | cap8 ≠ cap64 | **cap8 == cap64 (byte-identical)** |
| CoreMark CRCs | — | **unchanged** |
| `hoistLoopInvariantCode` hoists | 22 | **22** |
| `hoistLoopBoundsChecks` hoists | 5 | **5** |

(No regression in invariant-code or bounds-check hoisting — lazy synthesis only drops the *wasted* empty-preheader creation.)

## 2. `deadCodeElimination` ⇄ `deadLocalSetElimination` slow mutual convergence (keyvault monster)
As separate pipeline slots the two cleanups expose work for each other and micro-converge over many outer iterations (the keyvault monster still removed 1/3 insts at iter 7), re-running all ~25 passes each time. Clustered into a single `deadCodeAndLocalSetCleanup` slot that iterates them to their mutual local fixpoint, so the outer fixpoint settles fast. **Output-neutral at convergence** (verified: branch == baseline at a high outer cap).

## Validation
- `zig build test`: **87/87 steps, 4449/4456 passed (7 skipped)**.
- CoreMark CRCs identical; output now deterministic (cap-insensitive).

Refs #813.

### DIFF
--- a/src/compiler/ir/passes.zig
+++ b/src/compiler/ir/passes.zig
@@ -3830,6 +3830,7 @@ const pass_name_registry = [_]PassNameEntry{
     .{ .fn_ptr = &deadStoreElimination, .name = "deadStoreElimination" },
     .{ .fn_ptr = &deadCodeElimination, .name = "deadCodeElimination" },
     .{ .fn_ptr = &deadLocalSetElimination, .name = "deadLocalSetElimination" },
+    .{ .fn_ptr = &deadCodeAndLocalSetCleanup, .name = "deadCodeAndLocalSetCleanup" },
     .{ .fn_ptr = &hoistLoopBoundsChecks, .name = "hoistLoopBoundsChecks" },
     .{ .fn_ptr = &elideRedundantBoundsChecks, .name = "elideRedundantBoundsChecks" },
     .{ .fn_ptr = &foldLoadStoreOffset, .name = "foldLoadStoreOffset" },
@@ -4178,7 +4179,17 @@ pub fn hoistLoopInvariantCode(func: *ir.IrFunction, allocator: std.mem.Allocator
 
     var changed = false;
     for (lf.loops) |*loop| {
-        const ph = (try obtainLoopPreheader(func, loop, &predecessors, dom, allocator)) orelse continue;
+        // Defer preheader resolution until we actually hoist something.
+        // Synthesizing a preheader for a loop with nothing to hoist
+        // needlessly creates an empty `br header` block that
+        // `tailDuplicateSmallJoins` immediately absorbs again; the two
+        // passes then ping-pong forever (each iteration leaks one dead
+        // block) and the per-function fixpoint never converges, making
+        // codegen build-layout-sensitive (issue #813). Resolve — and, if
+        // necessary, synthesize — the preheader lazily on the first real
+        // hoist instead.
+        var ph_opt: ?ir.BlockId = null;
+        var ph_resolved = false;
         // One-shot scan of the loop body: which wasm-local / wasm-global
         // indices are written, and does the body contain any call or
         // memory-mutating op?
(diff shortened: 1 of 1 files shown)


cataggar (commented):
## Review: approve ✅

Reviewed in context and reproduced the key claims on a fresh worktree (aarch64 / Neoverse-N2, ReleaseFast).

### Correctness — both fixes sound
**Fix 1 (lazy preheader).** Root cause is precisely diagnosed: `hoistLoopInvariantCode` eagerly synthesized an empty `br header` preheader for *every* loop, which `tailDuplicateSmallJoins` (no min-body guard) re-absorbs → infinite ping-pong leaking one block/iteration. Deferring synthesis to the first real hoist breaks the cycle. Implementation details check out:
- `var block` + the re-fetch after synthesis (the `obtainLoopPreheader` call may append a block and reallocate `func.blocks`) correctly guards a real use-after-realloc hazard.
- `break :hoist_loop` on a null preheader is equivalent to the old `orelse continue`; the per-loop `defer` cleanup of `set_locals`/`set_globals` still runs.
- `hoistLoopInvariantCode` correctly stays in `passMutatesCfg` (it still mutates when it *does* hoist), so the dom/successor cache invalidation is preserved.

**Fix 2 (DCE⇄deadLocalSet cluster).** Correctly **not** registered in `passMutatesCfg` — verified neither sub-pass mutates the CFG, and the originals weren't listed either, so the dom-cache behavior is unchanged. The cap-16 inner loop is monotonic (only removes insts) and degrades gracefully if ever hit. All 8 production pipelines were updated consistently; the individual passes remain in the registry for tests/bisect.

### Verified independently
| check | result |
|---|---|
| `zig build test` | passes (exit 0) |
| CoreMark CRC (aarch64) | `crcfinal=0x33ff`, **Correct operation validated** |
| convergence (CoreMark `local_func=26`) | **main: max `iter=7` (capped) → this branch: `iter=3` (converges)** |

`iter=3 < cap 8` inherently proves the cap-insensitivity / determinism claim — it converges before any cap matters, so cap8 and cap64 are necessarily identical.

### Minor, non-blocking notes
1. **Bisect tooling nuance:** skipping `deadCodeElimination` / `deadLocalSetElimination` *by name* (`WAMR_AOT_SKIP_PASS`) no longer affects the production pipelines, since they now run *inside* `deadCodeAndLocalSetCleanup` (which calls them directly rather than as registered slots). Not a correctness concern; worth a note if pass-bisect relies on those names.
2. Lazy synthesis shifts *when* `predecessors` is updated relative to sibling loops — a theoretical ordering dependency, but the eager empty preheader had no lasting effect (it was absorbed anyway), and the byte-identical cap8==cap64 + full-suite results cover it.
3. x86_64 pipelines are changed symmetrically but weren't runtime-executed in my review (I'm on aarch64); the passing test suite covers the x86_64 codegen path.

Tightly scoped, correct, well-commented, and the central convergence claim reproduces exactly (7→3). LGTM.

Issue #818, opened 2026-06-07
perf(aarch64): fix loop-carried parallel_copy live-range gap and enable #540 phi-register residency (#808)


cataggar:
Child of #808 — the runtime lever with **measured headroom** on the CoreMark class.

On the N2 profile, the hot loops (`matrix_test` func7 ≈21%, `core_state_transition` func10 ≈25% of the run) spend roughly half their time round-tripping **loop-carried lowered-phi locals** through the frame (reload at the loop header, store at the back-edge). #540 routes phi resolution through register `parallel_copy` instead of frame slots, but it is **disabled in production**: `src/compiler/main.zig:364` → `if (false and target_arch == .aarch64)`, gated off a "CoreMark hang".

## Primary root cause (localized)
`computeLiveRangesImpl` (`src/compiler/ir/analysis.zig:1088`) records only the **first** def position and builds one positional `[start, end]` interval. A loop-carried `parallel_copy` dst is **defined on the back-edge predecessor** (late in global order) and **used at the loop header** (early). So `start = def_pos` (back-edge) and `end = max(start, last_use)` collapses to ≈`[backedge, backedge]` — the range **misses the header use**, and the allocator reuses/clobbers the loop-carried register → corruption / infinite loop (`matrix_test` hangs, `core_bench_list` wrong CRCs). Inner self-loops survive only because an entry-edge `parallel_copy` def yields an early `def_pos`.

### Fix
Treat `parallel_copy` dsts **phi-like** in range construction: pull `start` back to the earliest position where the value is live (its loop-header live-in), mirroring how `computeSsaLiveness` (`analysis.zig:332`) spans real phi dests across the back-edge. `computeLiveness` already marks the dst live-in at the header; the range builder must honor that for `start` (`start = min(def_pos, earliest_live_in_pos)`), not only extend `end`.

## Steps
1. Range fix in `computeLiveRangesImpl` + regression unit test: a loop-carried `parallel_copy` dst defined *after* its header use must get a range that spans the back-edge.
2. Flip the gate at `src/compiler/main.zig:364` (drop `false and`, keep aarch64-only).
3. Validate on the **aarch64 Neoverse-N2 self-hosted runner**:
   - functional: `matrix_test` (func7) no longer hangs; `core_bench_list` (func3) CRCs correct; func10 stays correct; full `zig build test` green.
[...]

Issue #540, opened 2026-05-15
perf(codegen): route phi-resolution through register MOV (#386 follow-up)


cataggar:
## Background

Current state: `lowerPhisToLocals` injects `local_set` / `local_get` pairs at every phi predecessor; aarch64 codegen lowers them to `frameStore` / `frameLoad` (see `aarch64/compile.zig:2122-2150` and `frameStore` / `frameLoad` helpers at `:2221` / `:2233`). Every phi arm therefore takes a memory round-trip.

The copy-hint plumbing landed in PRs #441 / #529 cannot elide these — it operates on register-to-register MOVs, and the phi resolution path produces no MOV at all.

## Proposed approach

Replace the per-arm memory round-trip with a **parallel-copy resolver** at each predecessor's terminator:

1. Collect all phi-arm `(src_reg → dst_reg)` pairs for the predecessor.
2. Detect cycles (need a temp register for swaps).
3. Emit a topologically-sorted MOV sequence; fall back to frame spill only on cycles without an available temp.
4. Spill to frame remains the safe fallback when registers conflict / no temp is free.

Once the phi resolution path produces MOVs, the copy-hint plumbing from PR #441/#529 will elide many of them automatically — measure the compound win, not just the parallel-copy fast path alone.

This is the **standard 4-phase parallel-copy algorithm**; naive sequential MOVs corrupt cyclic swaps like `a→b, b→a`.

## Touched files

- `src/compiler/codegen/aarch64/compile.zig` — primary (`lowerPhisToLocals` consumer; replace `frameStore`/`frameLoad` at lines 2122-2150 with parallel-copy resolver).

## Acceptance

- `scripts/bench_coremark.py --baseline origin/main --target HEAD --runs 10` aarch64 ≥ **+1 %**.
- `coremark.aot` byte size measurably down (memory round-trip → MOV is ~half the byte count).
- `scripts/bench_simd.py` no regression > 2 %.
- `zig build test` green.
- New unit tests: 2-arm phi with disjoint regs (MOV elided by copy-hint coalescer); 3-arm cyclic phi (`a→b, b→a` swap — needs temp); spill-fallback when no temp is free.

## Out of scope

- x86_64 phi resolution — file as a separate follow-up if aarch64 lands cleanly.

## Sequencing

[...]

PR #823, opened 2026-06-08
fix(aarch64): guard phi coalescing against lost-copy, enable #540 (#820)

### LINKED ISSUE #820: fix(aarch64): register-pressure codegen miscompile blocks #540 (core_bench_list/mergesort hang)
Child of #808. Split out from #818.

The #818 live-range fix (loop-carried `parallel_copy` start pull-back) is **necessary but not sufficient** to enable #540. With the gate flipped on, `core_bench_list` (inlined `core_list_mergesort`) still **hangs** on aarch64 (the merge never converges → `nmerges` stays > 1 → `insize` doubles to `0x80000000` → infinite outer loop).

## What it is NOT (ruled out with evidence)
- **The #818 range collapse.** A post-allocation conflict verifier using *true* per-position liveness (registers AND spill slots) finds **zero conflicts**, with and without the range fix. `analysis.computeLiveness` is pinned to a round-robin oracle by a differential test, so the interval set is genuinely conflict-free.
- **The #817 parallel-copy resolver.** A standalone simulation of the exact resolver algorithm over 200k randomized reg+stack batches (cycles, duplicate sources) passes with 0 failures.
- **Sub-passes.** Disabling the scheduler / range-split / move-coalesce / post-emit coalescer *individually* still hangs.
- **The lowering placement.** `coalescePhiLocalsToParallelCopy` places the `parallel_copy` at the same terminator the original `local_set` used; no critical-edge change.

## What it IS
A **register-pressure / register-assignment-dependent codegen miscompile** (the #109 `memory_fill` scratch-clobber class). Coalescing a **single** loop-counter phi (`synth local 77 → v533` in `core_bench_list`) is enough to hang in isolation. `v533` itself is allocated correctly (x0, range `[186,265]` covers its two live intervals; disassembly shows x0 is never clobbered on any main path between its defs and uses). Coalescing it reserves x0 across the merge loop and forces a *reallocation* of other vregs; the new (conflict-free) allocation then trips a latent instruction-level miscompile.

It is a **Heisenbug**: adding a single `volatile int` guard to `core_list_mergesort` makes it vanish (then `wamrc verify` matches wasmtime), so source instrumentation perturbs it away and `wamrc verify` cannot diff the unperturbed (hanging) build.

## Minimal reproducer
Build a standalone from the real CoreMark list sources (no full CoreMark needed):

```c
// lrepro.c
#include "coremark.h"
#include <stdio.h>
ee_u32 default_num_contexts = 1;
ee_u16 core_bench_state(ee_u32 b, ee_u8 *m, ee_s16 s1, ee_s16 s2, ee_s16 st, ee_u16 c){(void)b;(void)m;(void)s1;(void)s2;(void)st;return c;}
ee_u16 core_bench_matrix(mat_params *p, ee_s16 s, ee_u16 c){(void)p;(void)s;return c;}
static unsigned char memblk[140];
int main(void){
    core_results res; res.seed1=0; res.seed2=0; res.seed3=0x66; res.size=140;
    res.memblock[0]=(void*)memblk;
    res.list=core_list_init(res.size,(list_head*)memblk,res.seed1);
    ee_u16 crc=core_bench_list(&res,1); crc=core_bench_list(&res,-1);
    printf("crclist=%04x\n",(unsigned)crc); return 0;
}
```

```sh
cd tests/benchmarks/coremark
zig build-exe lrepro.c coremark/core_list_join.c coremark/core_util.c \
  -target wasm32-wasi -lc -O ReleaseFast -Icoremark -Icoremark/posix \
  -DSEED_METHOD=SEED_VOLATILE -DPERFORMANCE_RUN=1 -DHAS_PRINTF=1 -DITERATIONS=1 \
  --name lrepro -femit-bin=lrepro.wasm
# After temporarily flipping the #540 gate ON (src/compiler/main.zig):
../../../zig-out/bin/wamrc run lrepro.wasm   # HANGS (gate off: prints crclist=...)
```

A 5-node list (`memblk[140]`) hangs; ≤4 nodes are correct, so the corruption needs the merge to actually run.

## Suggested next step (non-perturbing)
Instruction-level diff of the `#540-on` vs `#540-off` `core_bench_list` disassembly across the first mergesort pass to find the first **logical** (not just register) divergence — or gdb store-by-store lockstep from the same input to find the first differing list-pointer store.

## Acceptance
- `wamrc verify` on the reproducer matches wasmtime with the #540 gate ON.
- Re-enabling the #540 gate (`src/compiler/main.zig`, drop `false and`) gives correct CoreMark CRCs and no hang; then #818 / #540 can land.

Refs #818, #808, #540, #817, #392.

cataggar:
Closes #820.

## Summary

Fixes the residual #820 miscompile that kept #540 (aarch64 phi→register `parallel_copy`) gated off, and enables #540.

`coalescePhiLocalsToParallelCopy` rewrites a phi (lowered to a synth local: `local_set` in each predecessor + `local_get` at the join) into a `parallel_copy dest <- src` at each predecessor terminator. For a loop-header phi this places the back-edge increment copy at the latch. In the original memory-slot form the join **reloads** the value on every header visit, so a use of the phi result reached *after* the loop is insulated from the latch store. Coalescing removes that reload, so the latch copy's value persists past the loop and an out-of-loop use observes the wrong (incremented) value — the classic **lost-copy / SSA-destruction** problem.

This hung CoreMark `core_list_mergesort` (inlined into `core_bench_list`): the merge's loop-counter phi was read by a `br_if` after the loop, so the branch saw the incremented counter, `nmerges` never reached 1, `insize` doubled to `0x80000000`, and the program looped forever.

## Fix

`synthCoalesceLostCopyUnsafe` drops any synth whose phi destination has a use **outside the natural loop** of the back-edge(s) into the join. Simple in-loop counters and forward/diamond phis (no back-edge) stay eligible — coalescing still fires for the common cases. With the hazard ruled out, the #540 gate is flipped on for aarch64 in `main.zig`.

## Validation

- Reproducer: `crclist=9218` (matches gate-off **and** wasmtime)
- `wamrc verify` reproducer: **OK vs wasmtime oracle**
- CoreMark CRCs identical to wasmtime (`crcfinal 0x65c5`, "Correct operation validated")
- Full suite: **4462/4474 passed** (12 skipped), incl. 2 new regression tests:
  - `coalescePhiLocalsToParallelCopy: rejects loop phi live across loop exit (#820)`
  - `coalescePhiLocalsToParallelCopy: coalesces loop phi used only inside the loop`

No regression possible below the gate-off baseline: unsafe phis fall back to the exact pre-#540 frame path; safe phis use register moves.

Builds on #818 (#821).

### DIFF
--- a/src/compiler/codegen/aarch64/compile.zig
+++ b/src/compiler/codegen/aarch64/compile.zig
@@ -9667,10 +9667,75 @@ test "compile: phi-resolution end-to-end produces no per-arm frame round-trip" {
     //   b0:  v_init = iconst 0; br b1
     //   b1:  v_i = phi (b0, v_init) (b1, v_next); v_next = v_i + 1;
     //        br_if cond → b1 else b2
-    //   b2:  ret v_i
-    // The phi-resolution should NOT emit `str ... [fp, ...]` and
-    // matching `ldr` purely for the phi bridge: instead a register MOV
-    // (or, when the coalescer wins, no instruction at all).
+    //   b2:  ret v_next
+    // The counter `v_i` is read only inside the loop (its post-loop result
+    // is observed via `v_next`, not `v_i`), so it is safe to coalesce and the
+    // phi-resolution should NOT emit `str ... [fp, ...]` and matching `ldr`
+    // purely for the phi bridge: instead a register MOV (or, when the
+    // coalescer wins, no instruction at all). A phi whose result is live
+    // across the loop exit takes the memory path instead (see the #820
+    // lost-copy tests above).
+    const allocator = std.testing.allocator;
+    var func = ir.IrFunction.init(allocator, 0, 1, 0);
+    defer func.deinit();
+
+    const b0 = try func.newBlock();
+    const b1 = try func.newBlock();
+    const b2 = try func.newBlock();
+
+    const v_init = func.newVReg();
+    const v_i = func.newVReg();
+    const v_one = func.newVReg();
+    const v_next = func.newVReg();
+    const v_cond = func.newVReg();
+
+    try func.getBlock(b0).append(.{ .op = .{ .iconst_32 = 0 }, .dest = v_init, .type = .i32 });
+    try func.getBlock(b0).append(.{ .op = .{ .br = b1 } });
+
+    const edges = try allocator.alloc(ir.Inst.PhiEdge, 2);
+    edges[0] = .{ .block = b0, .val = v_init };
+    edges[1] = .{ .block = b1, .val = v_next };
+    try func.getBlock(b1).append(.{ .op = .{ .phi = edges }, .dest = v_i, .type = .i32 });
+    try func.getBlock(b1).append(.{ .op = .{ .iconst_32 = 1 }, .dest = v_one, .type = .i32 });
+    try func.getBlock(b1).append(.{ .op = .{ .add = .{ .lhs = v_i, .rhs = v_one } }, .dest = v_next, .type = .i32 });
+    try func.getBlock(b1).append(.{ .op = .{ .iconst_32 = 10 }, .dest = v_cond, .type = .i32 });
+    try func.getBlock(b1).append(.{ .op = .{ .br_if = .{ .cond = v_cond, .then_block = b1, .else_block = b2 } } });
+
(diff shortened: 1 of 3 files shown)


cataggar:
Almost 17% faster on aarch64.

PR #839, opened 2026-06-11
wasi:http: block on a self-pipe waker for outbound HTTP futures (#833)

### LINKED ISSUE #833: wasi:http: pollable.block busy-spins on outbound HTTP futures (burns a full core; should block on a wakeable fd)
## Summary

`pollable.block` over a `future-incoming-response` (the `wasi:http/outgoing-handler.handle` path) doesn't actually block — the guest busy-spins `block → get → block` and pegs a full CPU core for the entire duration of the outbound HTTP round-trip. Wall-clock latency is fine (network-bound), but CPU usage is ~100× higher than it should be for an I/O-wait.

## Measurement

Same `wasi:http` component (the [`azure-sdk-for-zig` avs-wasi example](https://github.com/cataggar/azure-sdk-for-zig/tree/examples/arm_avs_wasi) — 3 paginated HTTPS GETs to `management.azure.com`), AOT-precompiled, macOS aarch64, wamr v3.0.0-dev.11:

| runtime | `real` (wall) | `user` (CPU) |
|---|---|---|
| **wamr (AOT)** | 1.70–2.21 s | **1.69–2.15 s**  ← `user ≈ real` |
| **wasmtime**   | 1.63–2.21 s | **0.01–0.02 s**  ← sleeps on the socket |

`/usr/bin/time -p` over 3 runs each. wamr spends essentially all of the wall-clock as on-CPU time; wasmtime spends ~10–20 ms of CPU and is otherwise asleep waiting on the network. The work is identical (same TLS handshakes, same REST calls, same JSON), so the delta is pure busy-wait.

## Root cause

In `src/component/wasi_cli_adapter.zig`:

- The outbound fetch runs on a worker thread (`httpFetchWorker`) that flips a shared `FutureIncomingResponse` from `.pending` to `.ready_ok` / `.ready_err`.
- `pollable.block` → `blockUntilAnyReady` → `waitForBackendEvents`:

  ```zig
  fn waitForBackendEvents(self, handles, timeout_ms) !bool {
      ...
      for (handles) |handle| {
          const pollable = self.lookupPollable(handle) ...;
          if (self.pollableBackend(pollable)) |backend| {
              try fds.append(..., .{ .fd = backend.fd, .events = backend.events, ... });
          }
      }
      if (fds.items.len == 0) return false;   // ← HTTP future has no backend fd
      _ = std.posix.poll(fds.items, timeout_ms) ...;
      return true;
  }
  ```

  The HTTP-future pollable has **no backend fd** (it's settled by a worker thread, not a socket the adapter owns) and **no timer deadline**, so `pollableBackend` returns null, `fds.items.len == 0`, and `waitForBackendEvents` returns `false` immediately. With no `earliest` timer either, `blockUntilAnyReady` returns at once. The guest transport then loops `pollable.block` → `future-incoming-response.get` (still pending) → `pollable.block` … with nothing ever putting the thread to sleep ⇒ a tight busy-poll until the worker finishes.

## Suggested fix

Give the HTTP-future pollable a **wakeable backend** so `waitForBackendEvents`'s `poll()` blocks until the worker signals completion. A self-pipe / eventfd works cross-platform:

- Allocate a pipe (or eventfd on Linux) per in-flight HTTP future.
- `httpFetchWorker` writes one byte after flipping the future to `.ready_*`.
- `pollableBackend` for an HTTP-future pollable returns `{ .fd = pipe_read_end, .events = POLLIN }`.
- Drain the byte when the future is observed ready.

Then `pollable.block` blocks in `poll()` until the round-trip completes (matching wasmtime's behaviour), and `user` CPU drops to ~the parse cost.

The same busy-wait likely affects any worker-thread-settled future without a backing fd (e.g. `future-trailers`); the fix should cover those uniformly.

## Notes

- Functionally correct today — this is purely a CPU-efficiency issue. On a laptop it's invisible; under concurrency (many simultaneous component invocations) it would burn a core per in-flight request.
- Not specific to the trampoline / arg-order fixes in #831; surfaced while benchmarking that work.

cataggar:
Closes #833.

## Problem
`pollable.block` over a `future-incoming-response` busy-spun a full core for the entire duration of an outbound HTTP round-trip (`user ≈ real`), where wasmtime spends ~0 CPU and sleeps on the socket. The outbound fetch runs on a worker thread that flips a shared `done` flag, but the `.http_future_response` pollable had **no backend fd** — so `waitForBackendEvents` saw zero fds, returned immediately, and the guest tight-looped `block → get(pending) → block`.

## Fix
Give worker-thread-settled HTTP futures a cross-platform self-pipe `Waker` on `PendingHttpFetchShared`:

- `httpFetchWorker` `signal()`s the waker right after `done.store(.release)` (via a top-level `defer`, covering every completion arm).
- `pollableBackend(.http_future_response)` returns the waker's `read_fd`, so the **P2** `blockUntilAnyReady → waitForBackendEvents` path parks in `poll()` until the worker completes.
- **P3** `driveAsyncEvents` now polls the gathered waker fds instead of waking on a fixed 1ms re-poll cadence.
- Wakers are `deinit()`'d (fds closed) in both drainers and both adapter-teardown loops — no leaks. One byte is written per fetch and the host never reads it (closes on settle), so the pipe never fills.
- Windows keeps its existing no-op poll path (pre-existing, out of scope).

Result: the host sleeps until the round-trip completes, matching wasmtime's CPU profile.

## Why not zio?
Evaluated [lalinsky/zio](https://github.com/lalinsky/zio) v0.14.0 and did **not** adopt it. zio's value is re-architecting an app's whole I/O model onto a cooperative coroutine event loop; getting real benefit here would require a wholesale rewrite of the adapter's entire I/O layer (sockets/streams/UDP/DNS/timers/poll) onto zio's loop, and it's an awkward fit since the adapter runs synchronously as host-function callbacks on the wasm execution thread, not inside a scheduler. #833 just needs "something for `poll()` to sleep on" — a self-pipe/eventfd does that in ~30 LOC and zio adds nothing over a raw fd for this. (Detail also posted on the issue.)

## Tests
- New `#833` regression test: an in-flight HTTP future exposes a backend fd and `poll()` parks until the worker signals (not a spin).
- Full suite passes: `zig build test --summary all` → all modules pass.

### DIFF
--- a/src/component/wasi_cli_adapter.zig
+++ b/src/component/wasi_cli_adapter.zig
@@ -7906,6 +7906,56 @@ pub const WasiCliAdapter = struct {
         return fh;
     }
 
+    /// Park in `poll()` on the self-pipe waker fds of in-flight HTTP
+    /// fetches until a worker signals its waker or `cap_ns.*` elapses,
+    /// instead of waking on a fixed re-poll cadence and burning CPU
+    /// (#833). Returns `true` if it parked on a waker poll (the caller
+    /// then skips its own sleep). A pending fetch lacking a waker
+    /// (pipe-init failure) shortens `cap_ns.*` to `http_repoll_ns` so
+    /// the caller's cadence fallback still observes it promptly.
+    ///
+    /// Windows has no `poll()` path (`std.posix.pollfd` does not exist
+    /// there), so the body is `comptime`-gated off on Windows — mirroring
+    /// `waitForBackendEvents` — and only the cadence shortening applies.
+    fn parkOnHttpWakers(
+        self: *WasiCliAdapter,
+        ci: *ComponentInstance,
+        cap_ns: *u64,
+        http_repoll_ns: u64,
+    ) bool {
+        if (comptime builtin.os.tag == .windows) {
+            const pending = self.pending_http_fetches.items.len > 0 or
+                self.pending_http_fetches_p3.items.len > 0;
+            if (pending) cap_ns.* = @min(cap_ns.*, http_repoll_ns);
+            return false;
+        }
+        var waker_fds: std.ArrayListUnmanaged(std.posix.pollfd) = .empty;
+        defer waker_fds.deinit(self.allocator);
+        var http_without_waker = false;
+        for (self.pending_http_fetches.items) |entry| {
+            if (entry.shared.waker) |w| {
+                waker_fds.append(self.allocator, .{ .fd = w.read_fd, .events = pollInEvents(), .revents = 0 }) catch {};
+            } else http_without_waker = true;
+        }
+        for (self.pending_http_fetches_p3.items) |entry| {
+            if (entry.ci != ci) continue;
+            if (entry.shared.waker) |w| {
+                waker_fds.append(self.allocator, .{ .fd = w.read_fd, .events = pollInEvents(), .revents = 0 }) catch {};
+            } else http_without_waker = true;
+        }
+        if (http_without_waker) cap_ns.* = @min(cap_ns.*, http_repoll_ns);
+        if (waker_fds.items.len == 0) return false;
+        // Block until any worker signals its waker or the cap (timer
+        // deadline / UDP re-poll budget / hint) elapses.
+        const timeout_ms: i32 = blk: {
+            const ms = (cap_ns.* +| 999_999) / 1_000_000;
+            if (ms > @as(u64, @intCast(std.math.maxInt(i32)))) break :blk std.math.maxInt(i32);
+            break :blk @intCast(ms);
+        };
+        _ = std.posix.poll(waker_fds.items, timeout_ms) catch {};
+        return true;
+    }
+
     /// `ComponentInstance.async_event_driver` hook (#551). Advances the
     /// host monotonic clock (honoring `monotonic_clock_override` so unit
     /// tests stay deterministic) and drains any due timer-futures into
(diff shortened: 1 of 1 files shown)


cataggar:
## Performance timing — busy-spin eliminated ✅

Reproduced the #833 measurement on `main` @ `c614a51` (this PR's merge). Same `avs-wasi` component (3 paginated HTTPS GETs to `management.azure.com`), AOT-precompiled, macOS aarch64, `/usr/bin/time -p`, 3 runs each.

| runtime | `real` (wall) | `user` (CPU) |
|---|---|---|
| **wamr (AOT) — before (from #833)** | 1.70–2.21 s | **1.69–2.15 s** ← `user ≈ real` (busy-spin) |
| **wamr (AOT) — after (this PR)** | 1.70–2.70 s | **0.08–0.10 s** ← sleeps on the socket |
| **wasmtime** | 1.78–2.44 s | **0.02 s** |

`user` CPU dropped from ~`real` (a pegged core for the entire round-trip) to ~0.1 s, now matching wasmtime's I/O-wait profile. The small residual over wasmtime is parse/JSON cost, not busy-wait. Output identical, exit 0.

> Methodology note: timed `wamr run` directly against a `wamrc compile-component` manifest rather than `wamrc run` — `wamrc run` spawns `wamr` as a child, so `/usr/bin/time` would only measure the `wamrc` parent's CPU, not the runtime's.

PR #884, opened 2026-07-11
jit: route lazily-compiled functions through JitCodeCache (#879)

cataggar:
## Summary

Fixes deferred item 5 of #879 (generalizing the #862 lazy-JIT spike): `LazyCompileDriver.compileFn` (`src/component/aot_compile.zig`) called `platform.mapExecutableCode` directly for each on-demand-compiled function, bypassing `JitCodeCache` entirely. This meant:

- `JitCodeCache.residentBytes()`/`mappingCount()` undercounted any module compiled with `lazy_jit`.
- A configured `WAMR_JIT_CODE_BUDGET_BYTES` didn't bound lazily-compiled code at all.

## Changes

- `JitCodeCache.register`/`unregister`/`checkBudget` are now `pub` (previously file-private to `runtime.zig`) so the compiler-side lazy-JIT driver in a different compilation unit can call them.
- `LazyCompileDriver.compileFn`: calls `JitCodeCache.checkBudget` before, and `JitCodeCache.register` after, its `platform.mapExecutableCode` call — mirroring the eager `mapCodeExecutable` path.
- `LazyJitState.free`: calls `JitCodeCache.unregister` for each lazily-compiled mapping's size alongside the existing `munmap`, so `AotInstance.destroy()` doesn't leave a residual registry count.
- New regression test (`src/tests/lazy_jit_spike_test.zig`) proving both the compile-time registration and the destroy()-time unregistration.
- Updated `docs/design/lazy-jit-spike.md`'s deferred-work list to mark this item done.

## Validation

`zig build test -Djit=true -Dlazy_jit=true --summary all` — full suite passes (23/23 test groups), including the new test.
`zig fmt --check` clean on all edited files.

## Related

- #857 (JitCodeCache itself)


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #891 (JitCodeCache integration for lazily-compiled mappings), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #885, opened 2026-07-11
jit: fix concurrent first-call race for the same lazy function (#879)

cataggar:
## Summary

Fixes deferred item 8 of #879 (generalizing the #862 lazy-JIT spike): `LazyJitState.compile_fn` had no synchronization, so two threads racing to be the first caller of the same still-pending lazy function on the same `AotInstance` could both invoke `compile_fn` concurrently.

Both racing compiles individually produce *correct* code (this isn't a wrong-answer bug), but the plain, unsynchronized writes to `compiled[local_idx]`/`pending[local_idx]` mean only the last writer's mapping survives being stored — every other racing thread's `platform.mapExecutableCode` mapping is silently **leaked** (never `munmap`'d by `LazyJitState.free`, since it's never stored anywhere) and its compile work wasted.

## Changes

- `LazyJitState` gains a `mutex: platform.Mutex` field (a spinlock — `std.Thread.Mutex` is IO-gated in this Zig version, matching the existing convention in `types.zig`/`sig_registry.zig`/`platform.zig`).
- `callFuncScalar`: unsynchronized `.acquire`-load fast-path peek at `pending[local_idx]` (cheap common case: already resolved), falling back to double-checked locking when the peek says "maybe still pending" — only one thread ever actually calls `compile_fn` for a given `local_idx`.
- Updated `LazyJitState`/`compile_fn` doc comments — the old caveat calling this an explicit unsafe gap no longer applies.
- New stress test: races 16 threads at the same still-pending function and asserts (via a counting shim wrapped around the real `compile_fn`) that the real compile happens **exactly once**, plus that every thread still gets the correct result.
- Updated `docs/design/lazy-jit-spike.md`'s deferred-work list to mark this item done.

## Validation

`zig build test -Djit=true -Dlazy_jit=true --summary all` — full suite passes (23/23 test groups). Ran the full suite **5 times in a row** to check for flakiness in the new concurrency stress test — no flakiness observed.
`zig fmt --check` clean on edited `.zig` files.

## Related

- #859 (thread-safety audit whose "configure before concurrent work" contract this closes a gap in)


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #894 (thread-safe per-slot atomic pending->compiling->ready state machine), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #886, opened 2026-07-11
jit: port lazy-compile spike (#862) to aarch64 (#879)

cataggar:
## Summary

Fixes deferred item 4 of #879 (generalizing the #862 lazy-JIT spike): the lazy_skip codegen-level plumbing was x86_64-only. Specifically:

- `aarch64/compile.zig`'s `CompileOptions` had no `lazy_skip` field.
- `compileCoreWasmCached`'s aarch64 dispatch branch never passed one.
- `PrecompileOptions.lazy_jit` explicitly rejected `target_arch == .aarch64` (`error.CoreCompileFailed`).
- `LazyCompileDriver.compileFn` hardcoded the x86_64 per-function codegen entry point — which would have silently executed **wrong-ISA machine code** had this ever been reached on an aarch64 host (a real correctness bug lying in wait, not just a missing feature).

## Changes

- `aarch64/compile.zig`: add `CompileOptions.lazy_skip` (mirrors x86_64's field) and wire it into `compileModuleCachedWithOptions`'s per-function loop to emit zero-byte placeholders for lazy-eligible functions. Add `compileFunctionForLazyJit`, a public per-function entry point (mirroring x86_64's `compileFunctionRAWithGlobalOffsetsPublic`) that drives the same real regalloc-based `compileFunctionImpl` the eager path uses.
- `aot_compile.zig`: wire `.lazy_skip` into the aarch64 dispatch branch, drop the `target_arch == .x86_64` guard, add `LazyJitOut.target_arch` so `LazyCompileDriver.compileFn` dispatches to the correct backend in a `switch` instead of assuming x86_64.
- `lazy_jit_spike_test.zig`: extend `can_exec_aot` to include aarch64 (it was still x86_64-only, so these tests silently self-skipped on any aarch64 build).
- `docs/design/lazy-jit-spike.md`: update Scope/Deferred-work sections.

## Validation

**Native x86_64** (primary target): `zig build test -Djit=true -Dlazy_jit=true` passes (23/23 groups), `zig fmt --check` clean on new code.

**aarch64** (no physical hardware available, matching #874's constraint): cross-compiled the full test suite for `aarch64-linux-musl` and ran it under `qemu-aarch64-static`. This exercises real generated aarch64 machine code end-to-end, not just a type-check:
- `lazy_jit_spike_test.zig`'s both tests pass, including the 200-function measured-effect test, which shows the same qualitative eager-vs-lazy compile-time reduction as the original x86_64 spike.
- 4486/4499 tests passed overall. The one failure (`coldstart_test`'s `noop.cwasm load+invoke under budget`) is a pre-existing, unrelated qemu-emulation-speed artifact — issue #395 already documents this exact budget being runner-speed-sensitive on macOS-arm64 CI, and qemu emulation is slower still. Confirmed this same test fails identically on `origin/main` cross-compiled the same way, before this change.

## Related

- #862 (original lazy-JIT design spike, x86_64-only)
- #874 (the sibling "no aarch64 hardware available" gap for W^X/MAP_JIT validation)


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #890 (aarch64 backend parity, validated on real Apple Silicon), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #895, opened 2026-07-11
jit: skip IR optimization passes for lazy functions too (#879)

cataggar:
## Summary

Fixes deferred item 6 of #879 (generalizing the #862 lazy-JIT spike): the original spike only skipped the final codegen step for deferred functions — `passes.runPassesWithOptions` still ran the full per-function pipeline (SSA promotion, GVN, LICM, redundant-load-forwarding, etc.) on every function eagerly, including deferred ones that would never be codegen'd this run. Per the design doc's own note, those passes — not codegen — dominate compile time for small/simple functions, so this was leaving most of the actual win on the table.

## Changes

- `passes.zig`: add `RunOptions.lazy_skip` (same indexing/semantics as the codegen-level field of the same name from #862). The per-function loop in `runPassesWithOptionsScoped` now skips SSA promotion, the pass fixpoint, `scrubUnreachableBlocks`, and their verify calls entirely for lazy-eligible functions, leaving their IR exactly as `frontend.lowerModule` produced it. Safe because codegen against un-optimized (but still lowered) IR is already a supported path (`-O0` skips this whole pipeline module-wide, and the existing correctness test already round-trips through `callFuncScalar` successfully).
- `aot_compile.zig`: thread the same `lazy_skip` slice `compileCoreWasmCached` already computes into the `passes.runPassesWithOptions` call.
- `lazy_jit_spike_test.zig`: new test with a constant-foldable fixture (`x + 2 + 3`) proving the deferred function's retained IR matches completely untouched raw lowering (not the folded/optimized form) — a deterministic, non-timing-based check.
- `docs/design/lazy-jit-spike.md`: mark this deferred-work item done.

## Validation

`zig build test -Djit=true -Dlazy_jit=true --summary all` — full suite passes (23/23 groups). `zig fmt --check` clean.

The existing 200-leaf-function measured-effect test's eager-vs-lazy gap widened dramatically now that passes are skipped too: this session measured `eager=1614290us lazy=38146us` per the test's own timing print (lazy ~2.4% of eager), vs. the original spike's ~20% codegen-only reduction (`eager=359923us lazy=289020us`). Consistent with the design doc's own note that GVN/LICM/load-forwarding, not codegen, dominate compile time for small functions.

## Related

- #860 (the CoreMark pass-timing analysis this doc references for "GVN/LICM/load-forwarding are the expensive ones")


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #892 (deferring the per-function IR-optimization pass pipeline), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #896, opened 2026-07-11
jit: native trampoline for table/ref.func-reachable lazy functions, phase 1 (#879)

cataggar:
## Summary

Addresses M4.6 of generalizing the #862 lazy-JIT spike: functions reachable via `call_indirect`/`ref.func`+`call_ref` were entirely excluded from lazy eligibility, because `func_addrs[i]`/a table's native backing must always be a genuinely callable address, and the spike had nothing to offer there for a not-yet-compiled function.

**This is phase 1 of 2**, deliberately split (see updated `docs/design/lazy-jit-spike.md`): build and independently prove the two genuinely new pieces — the relaxed eligibility analysis, and the trampoline machine code — before wiring them into the already-large, correctness-critical `func_addrs`/table-population logic in `mapCodeExecutable`. Phase 2 (the wiring) is explicit follow-up, not this PR.

## Changes

- **`lazy_jit.zig`**: new `findLazyEligibleWithTrampoline`, lifting rule 3 ("any element segment disqualifies everything") at the pure-analysis level. Still enforces leaf + never-a-direct-call-target; additionally admits functions reachable via an element segment and/or `ref.func`, flagging which ones need a real trampoline via a parallel `needs_trampoline` output. New unit tests. The existing `findLazyEligibleLeaves` and its only call site are completely untouched.
- **`lazy_call_trampoline.zig`** (new file): a dedicated native trampoline pool (x86_64, SysV + Win64) — not `TrampolinePool`-based, see its doc comment for why. Each stub:
  1. saves every argument-passing register untouched,
  2. calls a caller-supplied `(ctx, local_idx) -> address` dispatcher (baked in as immediates, no shared global state),
  3. restores every register,
  4. tail-jumps to the resolved address.

  Works transparently under both real-call and inherited-tail-call entry, since it never disturbs the return address already on the stack. Unlike `host_trampolines.TrampolinePool` (single process-wide "active pool"), every stub is fully self-contained, so multiple `AotInstance`s can use this concurrently on different threads.
- **`lazy_call_trampoline_test.zig`** (new file): standalone tests using hand-written native payloads (no real wasm compile involved), proving dispatch receives the correct `(ctx, local_idx)`, all six argument-passing registers survive the round trip, multiple independently-allocated slots resolve independently, and the pool's cap is enforced.

## Still open (phase 2, follow-up)

Wiring `findLazyEligibleWithTrampoline` + `lazy_call_trampoline.zig` into `compileCoreWasmCached`/`mapCodeExecutable`/`AotInstance` so a real `call_indirect`/`call_ref` actually exercises this end-to-end.

## Validation

`zig build test -Djit=true -Dlazy_jit=true --summary all` — full suite passes (24/24 groups, up from 23 — the new `lazy_call_trampoline_test.zig` group). Ran 3x to check for flakiness in the new hand-written-asm tests (none observed). `zig fmt --check` clean on every new/changed file.

## Related

- #857 / #858 / #859 (JitCodeCache / W^X hygiene / thread-safety — this trampoline pool follows the same mmap/jitWriteProtect conventions `TrampolinePool` already established)


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #888 (call_indirect/ref.func/call_ref via native trampolines), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #897, opened 2026-07-11
jit: wire native trampoline + relaxed eligibility end-to-end (#879 M4.6 phase 2)

cataggar:
## Summary

Completes M4.6 of generalizing the #862 lazy-JIT spike. **Depends on #896** (phase 1 — this branch is stacked on `879-lazy-jit-trampoline`, so this diff will show phase 1's commits too until #896 merges; only the last commit is new here).

Real `call_indirect`/`ref.func`+`call_ref` calls to a still-pending lazy function now compile it on demand via `lazy_call_trampoline.zig`'s native trampoline (built in #896), instead of that reachability disqualifying the function from lazy eligibility entirely.

## Changes

- **`aot_compile.zig`**: `compileCoreWasmCached` now calls `lazy_jit.findLazyEligibleWithTrampoline` instead of `findLazyEligibleLeaves`, and threads a new `needs_trampoline_indices` list through `LazyJitOut` alongside the existing `lazy_local_indices`. `LazyCompileDriver` gains a `trampoline_pool` (torn down in `deinit()`). `setupLazyJit` pre-allocates a trampoline stub for every `needs_trampoline` index and records its address on `inst.lazy_jit.trampoline_addrs`.
- **`runtime.zig`**: `LazyJitState` gains `trampoline_addrs` (populated by `setupLazyJit`) and a new `resolve()` method extracted from the previously-inline `callFuncScalar` hook, now shared with a new `callconv(.c)` `trampolineDispatch` entry point matching `lazy_call_trampoline.LazyDispatchFn`'s signature exactly — so a function reachable through both a direct export call and a `call_indirect`/`call_ref` path is still only ever compiled once. `mapCodeExecutable` installs `trampoline_addrs` into `func_addrs` right after building it for local functions, before that array feeds `funcptrs`/`ptr_to_sig`/every table's native backing — a single patch point is enough for all of them.
- **Ordering change** (documented in `setupLazyJit`'s doc comment): `setupLazyJit` must now run **before** `mapCodeExecutable`, so trampoline addresses are ready in time for the `func_addrs` patch above. Updated the one existing call site accordingly.
- **Fixed a latent bug**: `findLazyEligibleWithTrampoline`'s (and the original `findLazyEligibleLeaves`'s) leaf check tested for `.call`/`.call_indirect` but not `.call_ref` — surfaced by writing the first real `call_ref` test fixture, where a function whose only outgoing call was a `call_ref` was incorrectly treated as a leaf. Fixed in both functions plus the module doc comment; new regression test.
- **Two new end-to-end tests** (`lazy_jit_spike_test.zig`): a real `call_indirect` through a table to a still-pending lazy function, and a real `ref.func`+`call_ref` to one, both proving on-demand compilation and correct results — not just that the eligibility analysis marked them right.

## Validation

`zig build test -Djit=true -Dlazy_jit=true --summary all` — full suite passes (24/24 groups — same count as #896, since these are new tests in an existing group, not a new build step). Ran 3x to check for flakiness in the new end-to-end trampoline-through-real-compiled-code tests (none observed). `zig fmt --check` clean.

## Related

- #896 (phase 1 — trampoline pool + relaxed eligibility analysis, standalone)


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #888 (call_indirect/ref.func/call_ref via native trampolines), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #898, opened 2026-07-12
jit: thread lazy JIT through component compilation (#879 M4.8 phase 1)

cataggar:
## Summary

Phase 1 of generalizing the #862 lazy-JIT spike to components (M4.8). **Stacked on #897** (this branch is based on `879-lazy-jit-trampoline-wiring`, so the diff shows those commits too until #896/#897 merge; only the last commit is new here).

`precompileComponentInMemory` (used by `wamr`'s in-process component JIT path, #854) compiled each core module via the `compileCoreWasm` wrapper, whose always-empty `CompileCacheCtx` made `opts.lazy_jit = true` fail immediately on every core (`lazy_jit` requires `cache_ctx.lazy_jit_out != null`). So component-level lazy JIT was previously completely non-functional — this fixes the compile-layer plumbing.

## Changes

- **`aot_compile.zig`**: `precompileComponentInMemory` now calls `compileCoreWasmCached` directly and threads a per-core `lazy_jit_out` through when `opts.lazy_jit` is set. `InMemoryPrecompiled` gains `lazy_jit_outs` (one entry per core, `&.{}` when `lazy_jit` is off — zero behavior change for every existing non-lazy caller, confirmed by the full existing component test suite still passing unmodified). Ownership mirrors the single-core `LazyJitOut` contract: each entry is meant to be move-consumed by `setupLazyJit` per core.
- **New test** reusing the existing `h1-compose` fixture (2 core modules): proves each core's eligibility is computed independently — core 0's single leaf function is eligible, core 1's function (which calls core 0's export through a cross-instance import) correctly is not, same as a same-core call would disqualify it. Exercises core 0's `LazyJitOut` end-to-end through `setupLazyJit` + `mapCodeExecutable` + on-demand compile via `callFuncScalar`.

## Still open (phase 2, follow-up)

Wiring `lazy_jit_outs` into `component/instance.zig`'s per-core instantiation loop so a `-Djit=true` component actually defers anything end-to-end. Deliberately left separate: `instantiateWithOptions` is a large, correctness-critical function (AOT/interpreter fallback, section-aware core-instance walking) that deserves its own focused pass, mirroring how `mapCodeExecutable`'s wiring got its own phase in the call_indirect/ref.func work (#896/#897).

## Validation

`zig build test -Djit=true -Dlazy_jit=true --summary all` — full suite passes (24/24 groups). Ran 3x to check for flakiness (none observed). `zig fmt --check` clean.

## Related

- #896 / #897 (the call_indirect/ref.func trampoline work this mirrors the phase-split approach of)


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #889 (threading lazy-JIT through the in-memory component precompile path), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #899, opened 2026-07-12
jit: wire per-core LazyJitOut into component instantiation + CLI (#879 M4.8 phase 2)

cataggar:
## Summary

Completes M4.8 of generalizing the #862 lazy-JIT spike. **Stacked on #898** (this branch is based on `879-lazy-jit-components`, so the diff shows those commits too until #896/#897/#898 merge; only the last commit is new here).

A real component instantiated through `component/instance.zig`'s AOT core-instantiation path (the same path `wamr run`/`wamr serve` use) can now actually defer eligible functions, not just have them computed at the compile layer with nowhere to plug in.

## Changes

- **`core_backend.zig`**: new `LazyJitSetupHook` (type-erased, mirroring `LazyJitState.compile_fn`'s existing #695 boundary — this file must not depend on compiler types) and `PrecompiledCore.lazy_jit_setup`. New `Options.findLazyJitSetup` mirrors `findPrecompiled`'s exact `core_wasm`/`module_idx` matching rule.
- **`aot_compile.zig`**: `precompileComponentInMemory` populates `lazy_jit_setup` per core via new `setupLazyJitOpaque`/`deinitLazyJitDriverOpaque` bridge functions.
- **`instance.zig`**: the AOT core-instantiation block looks up each core's hook and calls it right before that core's `mapCodeExecutable`, storing the opaque driver handle on `CoreInstanceEntry` for cleanup after `aot_runtime.destroy`.
- **`main.zig`**: the component JIT path opts into `opts.lazy_jit` automatically whenever built with `-Dlazy_jit=true` — that build flag is already the deliberate opt-in, no separate runtime toggle needed.
- **New end-to-end test**: a real 2-core component instantiated via `instantiateWithOptions` (not a hand-driven single `AotInstance`) proves a cross-instance call compiles the callee on demand.

## A real bug caught along the way

Enabling this for real CLI usage immediately caught a genuine bug via the **pre-existing** `jit_component_smoke` build-time smoke test (`wamr run` against the real 3-core `stdio-echo.wasm` fixture): `LazyCompileDriver.deinit` freed `LazyJitOut`'s retained `lazy_local_indices`/`needs_trampoline_indices` with whatever allocator `setupLazyJit` itself was called with, not the allocator `compileCoreWasmCached` actually built them with. Harmless when a test passes the same allocator for both (every prior single-core test did), but a silent leak when they differ — which they routinely do for a real component instantiated into a per-request arena.

Fixed by giving `LazyJitOut` its own `allocator` field, used by `LazyCompileDriver.deinit`/`InMemoryPrecompiled.deinit` instead of `self.allocator`. Also fixed `InMemoryPrecompiled.deinit` to unconditionally clean up every `lazy_jit_outs` entry (both consumption paths now hollow the source slot to a deinit-safe placeholder afterward), so a core that falls back to the interpreter and never runs its hook still gets cleaned up.

## Validation

`zig build test -Djit=true -Dlazy_jit=true --summary all` — full suite passes (24/24 groups). Manually verified `echo hello | wamr run stdio-echo.wasm` produces zero `DebugAllocator` leak reports across repeated runs, both before finding the allocator bug (confirming it was real) and after the fix (confirming it's resolved). Ran the full suite 3x with an explicit leak-count check given the subtlety of the bug found. `zig fmt --check` clean.

## Related

- #898 (phase 1 — compile-layer per-core LazyJitOut plumbing)


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #889 (threading lazy-JIT through the in-memory component precompile path), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #901, opened 2026-07-13
jit: IR-rewrite calls to indirect ref_func+call_ref for non-leaf lazy JIT, phase 1 (#879)

cataggar:
## Summary

Part of #879 (generalizing the #862 lazy-JIT design spike). This is **M4.7 phase 1**: non-leaf functions, standalone mechanism only.

Stacked on #899 (M4.8 phase 2) — diff will shrink once earlier PRs in the stack merge.

## Problem

Non-leaf functions can't be lazy-eligible today because a direct `.call` is codegen'd as a compile-time PC-relative branch patch (`CallPatch`) against the target's offset within one contiguous code blob. A lazily-compiled function's machine code instead lives in its own, separately-`mmap`'d region (potentially gigabytes away, and not a compile-time-constant offset anyway) — a rel32 branch cannot reliably reach it.

## Approach

Rather than modifying the hot-path `.call` instruction-selection codegen directly (the most-exercised path in the whole compiler — high risk of regressing everything, not just lazy JIT), this adds `lazy_jit.rewriteLocalCallsToIndirect`: a standalone **IR-level transformation** that rewrites `.call{func_idx, args, extra_results, tail}` targeting a LOCAL function into an equivalent `.ref_func{func_idx}` + `.call_ref{type_idx, func_ref, args, extra_results, tail}` pair.

Both replacement instructions already exist and are already tested:
- `.ref_func` already reads `vmctx.funcptrs[idx]` — the same array `mapCodeExecutable` patches with either a function's real compiled address or a native trampoline address (#879 M4.6).
- `.call_ref` already dispatches through it via register-indirect `call`/`jmp` (not a PC-relative patch), with full support for tail calls, HRP/multi-result, and stack-passed args — the same call shapes `.call` needs.

Import calls (`func_idx < import_count`) are left untouched since their codegen already dispatches indirectly through `vmctx.host_functions[]`.

`type_idx` is threaded through for completeness but is purely informational at this codegen path — `.call_ref` targets are statically typed by validation (unlike `call_indirect`'s runtime table-entry signature check), so nothing depends on its exact value at dispatch time.

**Self-recursion is safe**: `funcptrs[F]` is read at call *execution* time, not compile time, and is updated to F's real address before F's own first-call compile returns control to any caller — so a recursive call inside F correctly resolves to F's real code by the time it actually executes.

## Testing

- Unit tests on hand-built IR proving the rewrite's structural correctness (local call → ref_func+call_ref, import calls left alone, tail flag/type_idx/args preserved, multiple calls in one block).
- Three new end-to-end tests in `lazy_jit_spike_test.zig`, gated behind a new opt-in `PrecompileOptions.force_indirect_calls` (applies the rewrite to every function in the module) that route through the **real** compile → emit → instantiate → execute pipeline:
  - a plain direct call,
  - a tail call (`return_call`, exercising the real-tail `.call_ref` codegen path),
  - a self-recursive call,

  each compared against the same wasm compiled the normal way (identical results).
- Full suite: `zig build test -Djit=true -Dlazy_jit=true` — 4511/4518 passed (7 pre-existing skips), 0 failures, run 3x for stability. `zig fmt --check` clean on all changed files.

## Scope / what's NOT here

Deliberately **not** wired into `lazy_jit`'s eligibility rules (`findLazyEligibleWithTrampoline`) or the lazy-compile path (`LazyCompileDriver.compileFn`) yet — that's a follow-up phase 2, mirroring the M4.6 (trampoline) and M4.8 (components) phase splits already used in this stack. This phase is purely the standalone, independently-testable rewrite mechanism.

docs/design/lazy-jit-spike.md updated with a phase-1-progress note under the "Non-leaf functions" deferred-work bullet.


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #887 (non-leaf x86_64 direct-call graphs -- via stable entry stubs instead of this PR's ref_func/call_ref IR-rewrite approach), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #902, opened 2026-07-13
jit: wire indirect-call rewrite into non-leaf lazy JIT eligibility (#879 M4.7 phase 2)

cataggar:
## Summary

Part of #879 (generalizing the #862 lazy-JIT design spike). This is **M4.7 phase 2**: wiring the standalone IR-rewrite mechanism from #901 (M4.7 phase 1) into the real eligibility rules and lazy-compile path.

Stacked on #901 (M4.7 phase 1) — diff will shrink once earlier PRs in the stack merge.

## Changes

- `LazyCompileDriver.compileFn` now runs `lazy_jit.rewriteLocalCallsToIndirect` on a function's IR before codegen, so any outgoing `.call` to a local function becomes a `.ref_func`/`.call_ref` pair that can reach a callee living in a different mmap'd region. A no-op for genuinely leaf functions (nothing to rewrite).
- `findLazyEligibleWithTrampoline`'s rule 1 relaxes from "leaf" to "no `call_indirect`": a function containing `.call`/`.call_ref` is no longer automatically disqualified (`.call_ref` was already indirect and safe; `.call` becomes safe via the rewrite above). `call_indirect` remains conservatively disqualifying — rewriting it is a separate, harder follow-up (a different dispatch/table mechanism entirely). Rule 2 ("never a direct call target") is deliberately left unrelaxed: relaxing it would require applying the same rewrite to the *eager* compile path too (every caller of a would-be-eligible target), a larger, separate, higher-risk change to the hot compile path affecting every module, not just lazy ones.

Net effect: exported "driver"-style functions that call other (local or imported) helpers, or perform `call_ref`, are now lazy-eligible as long as they're never themselves a direct `.call` target and never use `call_indirect`.

## Test fixture updates (expected, not regressions)

Several existing fixtures picked up newly-eligible functions as a direct consequence of the relaxed rule — each updated in place with the new expected counts and an explanation:

- The M4.6-phase-2 `ref.func`+`call_ref` fixture's `caller` function (previously disqualified purely for containing a `call_ref`) is eligible now. The test now asserts 2 pending indices instead of 1 and confirms both get resolved correctly.
- The M4.8 `h1-compose` fixture's core-1 function `g` (previously disqualified purely for containing a `.call` to a cross-instance import — even though import calls already dispatch indirectly through `vmctx.host_functions[]` and were never actually unsafe to defer) is eligible now too. Verified end-to-end via the existing real component-instantiation test: the cross-instance call into `g` now also exercises `g` itself being lazily compiled on first call (not just the `f` it calls), proving the existing M4.8 phase 2 lazy-compile hook generically supports newly-eligible non-leaf functions with no further changes needed.

## New/updated unit tests

In `lazy_jit.zig`:
- A function containing `.call` to a local function is now eligible (previously wasn't).
- A function containing `.call_indirect` still isn't eligible (unchanged).
- The pre-existing `call_ref`-containing-function test flips its expectation from ineligible to eligible, with an updated explanation.

## Testing

`zig build test -Djit=true -Dlazy_jit=true` — 4513/4520 passed (7 pre-existing skips), 0 failures, run 3x for stability (including in a fresh worktree after popping the change set, to make sure nothing was lost in transit). `zig fmt --check` clean on all changed files.

## Scope / what's still open

`docs/design/lazy-jit-spike.md` updated to mark M4.7 phase 2 done under the "Non-leaf functions" deferred-work bullet. Remaining for a future follow-up (not attempted here): relaxing rule 2 (would need eager-path call-site rewriting too) and supporting `call_indirect`-containing functions.


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #887 (non-leaf x86_64 direct-call graphs), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

PR #903, opened 2026-07-13
docs: M4.9 tiering design notes for lazy JIT (#879), no code changes

cataggar:
## Summary

Part of #879. This is **M4.9 (design notes only)** — per explicit direction, no code was written this session for tiering; this documents the investigation and findings for whoever picks it up next.

Stacked on #902 (M4.7 phase 2) — diff will shrink once earlier PRs in the stack merge.

## What this documents

While scoping M4.9 (tiering / background re-compilation to a higher optimization level), I found a real architectural snag: once a lazy function is first resolved, every subsequent call reads its native address directly out of `vmctx.funcptrs[]` and jumps — there's no Zig-level code left in the hot path at all (that's the whole point of the M4.6 trampoline design and the M4.7 `.call`→`.ref_func`+`.call_ref` rewrite). That also means there's no already-instrumented place left to count calls and decide when a function is "hot enough" to promote.

The new section in `docs/design/lazy-jit-spike.md` (under the existing "Tiering" deferred-work bullet) covers:

1. **Three options for getting a hotness signal back**, with tradeoffs:
   - Compiled-in call counters (a real codegen change, taxes every call forever)
   - Unconditional promotion after the first call (simplest, no threshold tuning, but promotes some functions that never pay back the cost)
   - Keep tier-1 dispatch indirect forever (undermines the M4.6/M4.7 work it would sit on top of)
2. **A separate, independent hazard**: safely hot-swapping a promoted function's code without unmapping memory some other thread might still be executing — recommends never freeing a superseded tier rather than building quiescence detection/reference counting.
3. **A recommended shape** for a future implementation (unconditional promotion + never-free-old-tier + a genuine background thread, not a synchronous recompile that would just relocate the cold-start latency this feature exists to avoid), explicitly flagged as a design sketch, not a specification.

## Not in scope here

No code, tests, or build changes — purely a documentation addition, per explicit user direction to scope M4.9 as design-only this session.


cataggar:
Closing as superseded: PR #900 (merged into main) independently implemented this same #879 sub-item as #893 (tiering/background re-compilation design spike, docs only), validated end-to-end on real hardware. Rebasing this branch would mean reconciling two competing/incompatible dispatch architectures rather than a mechanical conflict fix, so I'm closing rather than merging to avoid duplicated/competing mechanisms in the codebase. See #900 for the integrated implementation.

